PR #44, opened 2019-11-18
Support for more field types in structs, and easier usage/addition of genji field tags 

JustinJudd:
This PR contains two major changes and features:

1. Switched struct parsing to use go/types instead of go/ast. Doing this I was able to add transparent support for Type Aliases and Type Definitions.
2. Switched the way and location that genji tags are parsed. By parsing them earlier in the field review process, they can be used for decisions regarding the parsing and understanding of the field itself. I added an example of an ignore tag (could be skip or exclude...). This `ignore` tag allows a struct to contain fields that genji will ignore and not try to read or write from/to the database. This is really useful if you have a struct that contains unsupported field types, like []string for example.


asdine:
@JustinJudd I have released the [v0.3.0](https://github.com/asdine/genji/releases/tag/v0.3.0) of Genji. Basically, I have removed the support for the `pk` struct tags and change the way struct tags work to make them more like JSON:

```go
// before
MyField int `genji:"<option>"`

// after
MyField int `genji:"<alias>,<coma separated list of options>"`
```

To keep mirroring as much as possible the JSON package, I suggest using the `ignore` option as following:

```go
MyField int `genji:"-"`
```

WDYT?

JustinJudd:
Yeah, I think making it work like the std json library would ease onboarding for new users. I'll work to try to rebase some of the parsing and struct tag changes onto the new branch.

asdine:
Hey @JustinJudd, I have recently removed the code generation feature because with the new Document API it would take too long to implement and to maintain.
Code generation is a cool feature that I'd like to keep in Genji in the future but I believe it's way too early, I think it would be more useful once the low API will be more stable.
I have switched to a reflect based solution which is much more flexible and not really much slower.
I'm closing this PR as it's no longer compatible

Thank you :pray:

PR #49, opened 2019-12-14
Fixes Issue #46 to execute sql script

### LINKED ISSUE #46: Execute SQL script
The command-line tool must support reading from stdin:

```bash
# bolt
genji my.db < script.sql

# memory
genji < script.sql

# badger
genji --badger data < script.sql
```

JustinJudd:
Fixes #46

### DIFF
--- a/cmd/genji/shell/shell.go
+++ b/cmd/genji/shell/shell.go
@@ -92,6 +92,14 @@ func Run(opts *Options) error {
 		return err
 	}
 
+	ran, err := sh.runPipedInput()
+	if err != nil {
+		return err
+	}
+	if ran {
+		return nil
+	}
+
 	e := prompt.New(
 		sh.execute,
 		completer,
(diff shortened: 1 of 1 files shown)


cmd/genji/shell/shell.go
```diff
@@ -256,9 +257,35 @@ func (sh *Shell) getDB() (*genji.DB, error) {
 		os.Exit(2)
 	}
 
+	err = sh.runPipedInput()
```
asdine: I don't think we should run this here but rather on L93-94. Basically, there are two modes:
- if we find something on stdin: exec the query, display the output, stop the program
- else, we run the shell

```bash
$ echo "CREATE TABLE foo; INSERT INTO foo (a) VALUES (1); SELECT * FROM FOO" | genji 
{
  "a": 1
}
$ ...
```

cmd/genji/shell/shell.go
```diff
@@ -256,9 +257,35 @@ func (sh *Shell) getDB() (*genji.DB, error) {
...
+	m := stat.Mode()
+
+	if (m&os.ModeNamedPipe) == 0 /*cat a.txt| prog*/ && !m.IsRegular() /*prog < a.txt*/ { // No input from terminal
+		return nil
+	}
+	data, err := ioutil.ReadAll(os.Stdin)
+	if err != nil {
+		return fmt.Errorf("Unable to read piped input: %w", err)
+	}
+	err = sh.db.Exec(string(data))
```
asdine: Since data may contain a SELECT, we should use `sh.db.Query` instead:

```go
st, err := sh.db.Query(st)
if err != nil {
  ...
}
defer st.Close()
return record.IteratorToJSON(os.Stdout, st)
```

PR #59, opened 2020-02-26
Refactor NewValue and support type definitions of basic types

jpeletier:
This PR does the following:

* Refactor `document.NewValue()` merging it with `reflectValueToValue()` as both were very similar, and `NewValue()` would not recognize simple type definitions, such as `type myString string` as valid, while `reflectValueToValue()` would and created inconsistencies.
* Added an error type to wrap `ErrUnsupportedType` so this can be checked outside explicitly.
* Upgraded tests to check user types are covered.


asdine:
Great job, thank you :+1:

PR #60, opened 2020-02-27
Fix ordering by numeric values when there is no index

jpeletier:
This PR fixes a tricky bug that happens when running queries with an `ORDER BY` clause in which the field is a numeric value and there is no index for that field.

The bug happens because in `plan.go` in the the wrong byte-encoded representation of the value was used, and that does not tell if a number is bigger or smaller than another one. For example:

`0` is encoded as `[]byte{128}`
`100` is encoded as `[]byte{228}`
`200` is encoded as `[]byte{128, 200}`

Doing a pure byte comparison yields that `100` is greater than `200`, because `[]byte{228}` is considered greater than `[]byte{128, 200}` by `bytes.Compare()`.

To fix the issue, I exported the `index.EncodeFieldToIndexValue()` so the same function is used for generating the index and for ordering when there is no index.

I also changed tests in `select_test.go`, but only to put numeric values that would trigger the bug and to actually test ordering by numeric fields, with and without index; a case that was missing. The old tests passed without touching.

### DIFF
--- a/index/index.go
+++ b/index/index.go
@@ -135,7 +135,7 @@ func (i *ListIndex) Set(val document.Value, key []byte) error {
 		return err
 	}
 
-	v, err := encodeFieldToIndexValue(val)
+	v, err := EncodeFieldToIndexValue(val)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


asdine:
Good catch!

PR #61, opened 2020-02-27
added Value.Compare()

jpeletier:
`Value.Compare()` compares two values performing best-effort comparisons.
I wrote it to fix #60, but the final solution there turned out to be much better. In any case, I think this could be helpful enhancement, so here it is.


asdine:
Good idea :+1:

PR #79, opened 2020-04-12
Simple cli tool fixes 

james-storey:
This pull-request fixes a couple of issues I had when working with the genji cli tool.
I ran into a compile error at first, because of a mismatched type between the cli.Flag interface and the cli.BoolFlag type. Correcting that was simple and allowed the app to compile after making the change.

When working with the cli tool, I was expecting to find a way to exit or quit the cli from the interactive prompt like in other database cli tools. Having not found a method to quit the application from the prompt, I added the `.exit` command to give an easy way to quit the app.


asdine:
Thank you!

PR #83, opened 2020-05-19
Add support for UNSET clause

yaziine:
This PR adds support for `UNSET` clause.

`Table.ValidateConstraints` has been exposed because it's needed for validating the constraints when unsetting a field. 

Tests have been improved along the way as well.

### DIFF
--- a/sql/query/update.go
+++ b/sql/query/update.go
@@ -2,6 +2,7 @@ package query
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/asdine/genji/database"
 	"github.com/asdine/genji/document"
@@ -17,14 +18,14 @@ const updateBufferSize = 100
 type UpdateStmt struct {
 	TableName string
 
-	// Pairs is used along with the Set clause. It holds
+	// SetPairs is used along with the Set clause. It holds
 	// each field with its corresponding value that
 	// should be set in the document.
-	Pairs map[string]expr.Expr
+	SetPairs map[string]expr.Expr
 
-	// Fields is used along with the Unset clause. It holds
+	// UnsetFields is used along with the Unset clause. It holds
 	// each field that should be unset from the document.
-	Fields []string
+	UnsetFields []string
 
 	WhereExpr expr.Expr
 }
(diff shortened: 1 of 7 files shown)


sql/query/update.go
```diff
@@ -119,6 +121,58 @@ func (stmt UpdateStmt) Run(tx *database.Transaction, args []expr.Param) (Result,
...
+// unset executes the Unset clause.
+func (stmt UpdateStmt) unset(d *document.FieldBuffer, t *database.Table) error {
+	for _, f := range stmt.UnsetFields {
+		_, err := d.GetByField(f)
+		if err != nil {
+			// The field doesn't exist, we process the next one.
+			continue
+		}
+
+		// The only error eventually returned by Delete is ErrFieldNotFound,
```
asdine, line 160: This comment is no longer necessary

sql/query/update_test.go
```diff
@@ -36,13 +47,13 @@ func TestUpdateStmt(t *testing.T) {
...
 
-			err = db.Exec("CREATE TABLE test")
+			err = db.Exec("CREATE TABLE test (a text not null)")
 			require.NoError(t, err)
 			err = db.Exec("INSERT INTO test (a, b, c) VALUES ('foo1', 'bar1', 'baz1')")
 			require.NoError(t, err)
 			err = db.Exec("INSERT INTO test (a, b) VALUES ('foo2', 'bar2')")
 			require.NoError(t, err)
-			err = db.Exec("INSERT INTO test (d, e) VALUES ('foo3', 'bar3')")
+			err = db.Exec("INSERT INTO test (a, d, e) VALUES ('foo3', 'bar3', 'baz3')")
```
asdine, line 56: The fact that `a` was missing in this document was to make sure we don't get an error if we try to SET a field that doesn't exist. I think we should keep this kind of case to prevent regressions, wdyt?

yaziine: I think that we still have this test when we set `f` which doesn't exist in the document:
> {"SET / No cond / with missing field", "UPDATE test SET f = 'boo'", false, `[{"a":"foo1","b":"bar1","c":"baz1","f":"boo"},{"a":"foo2","b":"bar2","f":"boo"},{"a":"foo3","d":"bar3","e":"baz3","f":"boo"}]`, nil}

Does it make sense?

asdine: Yes, but not with the `WHERE` clause: `WHERE a = ?`, `WHERE a = f`

yaziine: There is a test with the `WHERE` clause actually:
> {"SET / With cond / with missing field", "UPDATE test SET f = 'boo' WHERE d = 'bar3'", false, `[{"a":"foo1","b":"bar1","c":"baz1"},{"a":"foo2","b":"bar2"},{"a":"foo3","d":"bar3","e":"baz3","f":"boo"}]`, nil},

asdine: Indeed 👍

PR #86, opened 2020-05-28
Add support of missing types for CREATE statement

### LINKED ISSUE #85: Cannot create table with array type
I would like to create a table with an array type but I have an error:
```sql
genji> CREATE TABLE foo (name text primary key, animals array);
found array, expected ) at line 1, char 50
genji>
```

Is it by design or not?

yaziine:
This adds the support of missing types for the `CREATE` statement.

I added tests that assert the entire set of types.

It fixes #85.


yaziine:
All I can do, I guess, is to add more type constraint [here](https://github.com/genjidb/genji/blob/master/sql/query/create_test.go#L52). But since that this PR focuses only on the `parser` package, adding tests for this one is enough I think.

We are pretty confident that the execution works well, the only missing part was parsing correctly those types.
What do you think?

asdine:
Actually we are not, because nothing is telling us trying to insert into a table with an array or document field constraint would work.
This PR pretty much completes what was missing for the field constraints support and constraint validation, it makes optional schemas more solid and complete. All it needs to be absolutely complete is tests that make sure everything executes well when inserting data.
With tests, we can leave that feature behind and move on with confidence to next ones. Wdyt?

sql/query/insert_test.go
```diff
@@ -128,4 +128,53 @@ func TestInsertStmt(t *testing.T) {
 		require.NoError(t, err)
 		require.JSONEq(t, `{"a": "a", "b-b": "b"}`, buf.String())
 	})
+
+	t.Run("with types constraints", func(t *testing.T) {
```
asdine, line 132: This is great! There is only one thing missing: making sure when we insert something that doesn't follow the constraint, we have an error. For constraints with and without `NOT NULL`.

yaziine: Good call! I've added a test for this in 27a1b2a. I'm wondering if it's enough or should I add a test for each constraint type. What do you think?

asdine: I think you can easily create a table-driven test for each type, each tech would only need to replace the type in the query. ex: `db.Exec("CREATE TABLE foo (a "+ test.FieldConstraint +")")`

yaziine: I pushed more test in 300447e.

PR #87, opened 2020-05-28
Introducing the query planner

### LINKED ISSUE #88: Building a query planner
Currently, there is one [separate type](https://github.com/genjidb/genji/blob/3dbc0f9c906524df393101ae7008b3e8c576393c/sql/query/select.go#L14) for each kind of statement, containing the entire logic of that statement. The parser only needs to determine which type to instantiate and give all the necessary information to that object so that it can be run.

There are several problems with this approach:
- SELECT, UPDATE, etc. share some logic (reading from a table, filtering, etc), which is currently duplicated
- Optimizations for one statement (indexes, etc.) are hard to apply to the other one
- The way the current "query planner" is designed makes it really hard to write optimizations.
- Adding nested queries or JOINS to the current code will make everything really complex

Having a better designed base would allow us to:
- build more complex queries:
   - GROUP BY
   - Nested queries
   - JOINs 
- share optimizations between queries

## Proposal

To solve that, we can build an intermediary representation of a query in the form of a tree.
A tree represents a stream of documents, from an *input* (table, index, etc) to an *output* (result stream, deletion from the table, update, etc.).
 
Each node of the tree represents an **operation** that will transform the stream (filter documents, project fields, renaming fields, etc.). These operations will follow some principles of relational algebra [1] [2].

With this representation, the difference between a SELECT and an UPDATE is only about changing one node of the tree.

The tree would be built by the parser before being passed to the execution phase.

Once we have this tree, we can optimize it using **Rule-based optimization**: Each rule is a function that takes one tree and returns a new tree. Rule-based optimization can be summarized like this:

```
t is a tree
r1, r2, ... rN are functions with this pseudo-signature r(t tree) tree

t -> r1() -> r2() -> ... -> rN() -> finalTree
```

Each rule does only one optimization:
- simplifying expressions
- determining if we can use indexes
- removing duplicate nodes
- ...

Once the tree is optimized, we can transform it into an actual [`document.Stream`](https://pkg.go.dev/github.com/asdine/genji@v0.5.0/document?tab=doc#Stream) that can be executed.

Some reads about relational algebra:
[1] http://infolab.stanford.edu/~ullman/fcdb/aut07/slides/ra.pdf
[2] https://www.geeksforgeeks.org/introduction-of-relational-algebra-in-dbms/

asdine:
This PR introduces a new `planner` package. This package will be used by the `parser` package to build a representation of the SELECT, UPDATE, and DELETE queries, using relational algebra and expression trees.
This tree will then be optimized by the query planner and then used to execute a query.

This fixes #88


sql/planner/tree.go
```diff
@@ -0,0 +1,509 @@
...
+
+var _ operationNode = (*selectionNode)(nil)
+
+// NewSelectionNode creates a node that filters documents of a stream, according to
+// the expression condition.
+func NewSelectionNode(n Node, cond expr.Expr) Node {
+	return &selectionNode{
+		node: node{
+			op:   Selection,
+			left: n,
```
yaziine, line 215: Is there any specific reason to set the left node?

sql/planner/tree.go
```diff
@@ -0,0 +1,509 @@
...
+	return
+}
+
+func (n *limitNode) toStream(st document.Stream) (document.Stream, error) {
+	stack := expr.EvalStack{
+		Tx:     n.tx,
+		Params: n.params,
+	}
+
+	v, err := n.limitExpr.Eval(stack)
```
yaziine, line 297: We should check that the `limitExpr` is not nil before evaluating it.

sql/planner/tree.go
asdine: Since we currently only read from one table, we only use one branch of the tree, that's why I only used the left branch of each node

sql/planner/tree.go
asdine: It's already checked before hand, the node cannot be created if the limitExpr is nil. I'm removing the limitExpr in another branch anyway 👍

asdine: I have cherry-picked the commit and added it to this PR

sql/planner/tree.go
```diff
@@ -0,0 +1,475 @@
...
+	cond   expr.Expr
+	tx     *database.Transaction
+	params []expr.Param
+}
+
+var _ operationNode = (*selectionNode)(nil)
+
+// NewSelectionNode creates a node that filters documents of a stream, according to
+// the expression condition.
+func NewSelectionNode(n Node, cond expr.Expr) Node {
```
yaziine, line 211: I think that for each `New*Node` we should check that the given Node and the remaining parameters (depending on the operation) are not null. Wdyt?

sql/planner/projection.go
```diff
@@ -0,0 +1,209 @@
...
+	cfg *database.TableConfig
+	tx  *database.Transaction
+}
+
+var _ outputNode = (*ProjectionNode)(nil)
+var _ operationNode = (*ProjectionNode)(nil)
+
+// NewProjectionNode creates a node that uses the given expressions to create a new document
+// for each document of the stream. Each expression can extract fields from the incoming
+// document, call functions, execute arithmetic operations. etc.
```
yaziine, line 30: I think the following is enough. It was redundant with the `ProjectionNode` documentation.
```suggestion
// NewProjectionNode creates a ProjectionNode.
```

sql/planner/projection.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return true
+}
+
+// Bind database resources to this node.
+func (n *ProjectionNode) Bind(tx *database.Transaction, params []expr.Param) (err error) {
+	n.tx = tx
+	if n.tableName == "" {
+		return
```
yaziine, line 87: Could we project without a table? Maybe we need to return an error here?

sql/planner/optimizer.go
```diff
@@ -0,0 +1,431 @@
...
+// - implements the indexIteratorOperator interface
+// - one of its operands is a field selector that is indexed
+// - the other operand is a literal value or a parameter
+// If found, it will replace the input node by an indexInputNode using this index.
+func UseIndexBasedOnSelectionNodeRule(t *Tree) (*Tree, error) {
+	n := t.Root
+	var prev Node
+	var inputNode Node
+
+	// first we lookup for the input node
```
yaziine, line 263: Shouldn't the `Input` node be the first in the tree?

sql/parser/select.go
```diff
@@ -1,65 +1,70 @@
...
 	// Parse "FROM".
 	var found bool
-	stmt.TableName, found, err = p.parseFrom()
-	if err != nil || !found {
-		return stmt, err
+	cfg.TableName, found, err = p.parseFrom()
+	if err != nil {
+		return nil, err
+	}
+	if !found {
```
yaziine, line 29: If the table is not found, shouldn't we return an error instead?

sql/query/expr/arithmeric.go
```diff
@@ -119,7 +131,7 @@ type bitwiseOrOp struct {
 
 // BitwiseOr creates an expression thats evaluates to the result of a & b.
```
yaziine, line 132:
```suggestion
// BitwiseOr creates an expression thats evaluates to the result of a | b.
```

sql/query/expr/arithmeric.go
```diff
@@ -137,7 +149,7 @@ type bitwiseXorOp struct {
 
 // BitwiseXor creates an expression thats evaluates to the result of a & b.
```
yaziine, line 150:
```suggestion
// BitwiseXor creates an expression thats evaluates to the result of a ^ b.
```

sql/planner/tree.go
asdine: I don't think the check is necessary as it is meant to be used internally. For now, we'd rather have a panic if it happens and rely on tests.

sql/parser/select.go
asdine: No because we can do this: `SELECT 1 + 1`

sql/planner/projection.go
asdine: Yes we can, `SELECT 1 + 1` gives this `∏(1 + 1)`

sql/query/expr/expr.go
```diff
@@ -17,6 +17,26 @@ type Expr interface {
 	Eval(EvalStack) (document.Value, error)
 }
 
+type isEqualer interface {
+	IsEqual(Expr) bool
+}
+
+// Equal reports whether a and b are equal by first calling
```
asdine, line 24:
```suggestion
// Equal reports whether a and b are equal by first calling IsEqual
```

sql/planner/optimizer.go
asdine: It should be last because there can be multiple inputs (this is not yet supported) but only one output. If you take a look at the relational algebra tree examples, the inputs are always at the bottom, not at the root.

sql/planner/optimizer.go
```diff
@@ -0,0 +1,431 @@
...
+
+		n = n.Left()
+	}
+
+	if inputNode == nil {
+		return t, nil
+	}
+
+	// then we get the table indexes. here we will assume that at this point
+	// inputNodes can only be instances of tableInputNode.
```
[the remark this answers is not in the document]
asdine: This is not an error, I meant to use inputNodes in general

sql/parser/select.go
yaziine: Why could we do that?

asdine: This is standard, you can try with sqlite3 for example. This is a way to execute an expression and get the result

PR #92, opened 2020-06-07
Allow dot notation for UPDATE SET 

### LINKED ISSUE #84: Support dot notation on UPDATE ... SET

```sql
```

- `a` is an array

### Exception for arrays

```sql
UPDATE users SET a.2 = 1;
{"a": [1, 0, 1]}
```

tzzed:
This is a proposal to fix #84

Fixes #84 for arrays:

```sql
UPDATE users SET a.2 = 1;
{
  "a": [1, 0, 1]
}
{
  "a": [2, 0]
}
```

The field changes correctly and the index is not created if it doesn't exist.

Fixes #84 for documents:

```sql
INSERT INTO users (name, age, nen, parents, abilities) VALUES
    ('Gon', 13, 'Enhancement', {'father': 'Ging Freecs'}, ['Jajanken']),
    (
        'Kirua', 14, 'Transmutation',
        {'father': 'Silva Zoldyck', 'mother': 'Kikyo Zoldyck'},
        ['Lighning Palm', 'Thunderbolt', 'Godspeed']
    );

UPDATE users SET parents.mother = "Brigit Zoldyck" where name = "Kirua";

SELECT * FROM users WHERE name = "Kirua";
{
  "name": "Kirua",
  "age": 14,
  "nen": "Transmutation",
  "parents": {
    "father": "Silva Zoldyck",
    "mother": "Brigit Zoldyck"
  },
  "abilities": [
    "Lighning Palm",
    "Thunderbolt",
    "Godspeed"
  ]
}
```

The fields change correctly. 
The field is created if it doesn't exist.

I did not units test because of I cannot compare arrays.


document/document.go
```diff
@@ -67,15 +67,78 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+		}
+		buf = buf.Append(va)
+		return nil
+	})
+
+	return buf, err
+}
+
+// SetDotNotation allow dot notation and replace value at the given index.
+func (fb *FieldBuffer) SetDotNotation(fname string, v Value) {
```
asdine, line 86: I think this logic could be written in the [`Set`](https://github.com/genjidb/genji/blob/b4c31313424dfe91b04a3118225ce1a39a8322a1/document/document.go#L220) method. Basically, the `Set` method is what was used by the UPDATE logic, it is used to only support top-level field modification.
Since your PR changes that, I think modifying `Set` rather than creating an alternative method is the way to go. What do you think?

asdine: Regarding the signature of the `Set` method, I think we should explicitly expect a `ValuePath` instead of a string. That way we don't have to cast fname to vpath, or to use the `isDotPath` function.

document/document.go
```diff
@@ -67,15 +67,78 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+	vPath := NewValuePath(fname)
+	index, strconvErr := strconv.Atoi(vPath[1])
```
asdine, line 88: I have trouble understanding `vPath[1]`, could you explain your reasoning?

This seems to be working for `foo.4` but not for `foo.bar.4` or `foo.bar.4.baz.1.2.3`, right?

document/document.go
```diff
@@ -67,15 +67,78 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+			var buf FieldBuffer
+			var b1 *FieldBuffer
+			// Cannot make field.integer on a document
+			if strconvErr == nil {
+				return
+			}
+
+			err := buf.Copy(f.Value.V.(Document))
+			if err != nil {
+				return
```
asdine, line 104: In Go, if a function can fail the error must be returned so that the caller can make a decision about that. 
If `buf.Copy` fails, this entire function will fail, but as it is, this function doesn't indicate if the set worked or not.
You should simply return the error with `return err`

sql/parser/select.go
```diff
@@ -63,7 +64,13 @@ func (p *Parser) parseResultFields() ([]query.ResultField, error) {
...
 	// Parse remaining (optional) result fields.
 	for {
-		if tok, _, _ := p.ScanIgnoreWhitespace(); tok != scanner.COMMA {
+		tok, _, _ := p.ScanIgnoreWhitespace()
+		if tok != scanner.COMMA {
+			p.Unscan()
+			return rfields, nil
+		}
+		tok, _, _ = p.ScanIgnoreWhitespace()
+		if tok != scanner.DOT {
```
asdine, line 73: This would allow users to write `SELECT foo, . foo FROM table`, could you explain your reasoning behind that?

sql/parser/update.go
```diff
@@ -62,6 +64,15 @@ func (p *Parser) parseSetClause() (map[string]expr.Expr, error) {
 			return nil, newParseError(scanner.Tokstr(tok, lit), []string{"identifier"}, pos)
 		}
 
+		p.Unscan()
+		ref, err := p.parseFieldRef()
+		if err != nil {
+			return nil, newParseError(scanner.Tokstr(tok, lit), []string{"identifier"}, pos)
+		}
+		if len(ref) > 1 {
```
asdine, line 72: Since we are dealing with `ValuePath`s now, I think the `parseSetClause` method should return something that takes that into account.
Otherwise, because of this `if`, we'd basically be doing this:

       Input                    Parsing                     Joining                  
     "foo.1.bar.0.1"  --> ["foo", "1', "bar", "0", "1"] --> "foo.1.bar.0.1"

document/document.go
```diff
@@ -67,15 +67,78 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
 	return Value{}, ErrFieldNotFound
 }
 
+// treatArrayValue update the value of array at the given index.
+func treatArrayValue(vlist Array, v Value, index int) (ValueBuffer, error) {
```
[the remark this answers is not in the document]
tzzed: It is ok for me, setArrayValue is more explicit.

document/document.go
tzzed: I discover `ValuePath ` and it  was a reflexion about change but I  hesitate to change the existing code.I am agree with you, I think `Set` method can be change and return an error if we do the `buf.Copy(..)` inside.

document/document.go
tzzed: The `vPAth[1]` is the result of simplest unique case `foo.4` but not `foo.bar.4.baz.1.2.3`. Must be change to manage the full dot notation.

document/document.go
tzzed: Totally agree. I will correct it. Thank you

yaziine:
@tzzed 
> A new PR coming soon.

Instead of creating a new PR, you can directly commit to this one.

asdine:
I can see that this PR as been continued in #113, closing this one

PR #98, opened 2020-06-12
Consider empty documents and arrays as falsy

### LINKED ISSUE #89: Inserting empty array and document into boolean field
If I insert an empty array or an empty document into a boolean field, I would expect that the value of that field will be `false`. Exactly the same way as with empty strings.

What do you think?

```sql
genji> CREATE TABLE test(a bool, b bool, c bool);
genji> INSERT INTO test VALUES {a: "", b: [], c: {}};
genji> SELECT * FROM test;
{
  "a": false,
  "b": true,
  "c": true
}
```

yaziine:
This fixes #89.

There is some significative changes in this PR:
- I added a `Length` function that returns the length of a document same as `ArrayLength` for arrays
- `Value.IsZeroValue` now returns an error next to the initial returned parameter because the underlying calls might return errors
- `Value.IsTruthy` now returns an error because of the previous point

### DIFF
--- a/document/document.go
+++ b/document/document.go
@@ -27,6 +27,20 @@ type Keyer interface {
 	Key() []byte
 }
 
+// Length returns the length of a document.
+func Length(d Document) (int, error) {
+	if fb, ok := d.(FieldBuffer); ok {
+		return fb.Len(), nil
+	}
+
+	var len int
+	err := d.Iterate(func(_ string, _ Value) error {
+		len++
+		return nil
+	})
+	return len, err
+}
+
 // FieldBuffer stores a group of fields in memory. It implements the Document interface.
 type FieldBuffer struct {
 	fields []fieldValue
--- a/document/document.go
+++ b/document/document.go
@@ -29,7 +29,7 @@ type Keyer interface {
 
 // Length returns the length of a document.
 func Length(d Document) (int, error) {
-	if fb, ok := d.(FieldBuffer); ok {
+	if fb, ok := d.(*FieldBuffer); ok {
 		return fb.Len(), nil
 	}
 
--- a/document/document.go
+++ b/document/document.go
@@ -19,9 +19,6 @@ type Document interface {
 	// GetByField returns a value by field name.
 	// Must return ErrFieldNotFound if the field doesnt exist.
 	GetByField(field string) (Value, error)
-	// IsEmpty returns true if the document is empty.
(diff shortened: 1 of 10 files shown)


document/document.go
```diff
@@ -27,6 +27,20 @@ type Keyer interface {
 	Key() []byte
 }
 
+// Length returns the length of a document.
+func Length(d Document) (int, error) {
+	if fb, ok := d.(FieldBuffer); ok {
```
asdine, line 32: `FieldBuffer` is usually a pointer, this optimization won't work as-is, we should infer for the pointer instead
```suggestion
	if fb, ok := d.(*FieldBuffer); ok {
```

document/value.go
```diff
@@ -494,27 +496,41 @@ func (v Value) ConvertToDuration() (time.Duration, error) {
...
-		return v.V == int64ZeroValue.V
+		isZeroValue = v.V == int64ZeroValue.V
 	case Float64Value:
-		return v.V == float64ZeroValue.V
+		isZeroValue = v.V == float64ZeroValue.V
 	case DurationValue:
-		return v.V == durationZeroValue.V
+		isZeroValue = v.V == durationZeroValue.V
+	case ArrayValue:
+		len, err := ArrayLength(v.V.(Array))
```
asdine, line 520: `Length` and `ArrayLength` are not optimal for this use case. Instead of counting all the values then comparing with 0, we only need to assert that there is at least one value. O(n) -> O(1)

yaziine: Very good call! I improved this in b34ff32. Please notice the `IsEmpty` method in the `Document` interface.

document/document.go
```diff
@@ -19,6 +19,9 @@ type Document interface {
 	// GetByField returns a value by field name.
 	// Must return ErrFieldNotFound if the field doesnt exist.
 	GetByField(field string) (Value, error)
+	// IsEmpty returns true if the document is empty.
+	// Otherwise, it returns false.
+	IsEmpty() bool
```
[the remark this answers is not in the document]
yaziine: Yeah actually I wasn't sure of that idea but I didn't find a better way.

"iterate only once" is a good hint actually, I will try something else.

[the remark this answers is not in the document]
yaziine: I removed the `IsEmpty` function and used iteration in 2143cb9.

Issue #100, opened 2020-06-16
Add .indexes command


yaziine:
List all indexes:
```
genji> .indexes
```

List indexes of a specific table:
```
genji> .indexes user
```

tzzed:
The behaviour of `.indexes` command should only list indexes name or returns table index information.
What kind of output is expected? Output json document or [Sql](https://dev.mysql.com/doc/refman/8.0/en/show-index.html)?

asdine:
Good question @tzzed, I think it should be the same as with the `.tables` command: Just a list of names, one per line

yaziine:
I think adding the name of the table and the field whose the index is configured to may worth it. 
```
genji> .indexes
idx_bar on user(name)
idx_foo on user(email)
```
where `user` is the table and `name` and `email` fields of that table.

asdine:
I love this idea, let's do it ❤️

tzzed:
What about the refactor of `Index struct`. Your suggestion leaves this issue out of context about the refactor.
So should I do an issue even to suggest the why?

PR #101, opened 2020-06-17
Dropping a table should remove only indexes associated to that table

### LINKED ISSUE #99: Dropping a table removes **all** indexes
Given the following queries:
```
On-disk database using BoltDB engine at path my.db.
genji> create table user(name text);
genji> create index idx_user_name on user(name);
genji> create table cars(brand text);
genji> create index idx_cars_brand on cars(brand);
```

Using `boltbrowser` I can see the indexes:

image

If I remove the `cars` table:
```
genji> drop table cars;
```

⚠️    **All indexes get removed**:

image

This is critical and should be fixed as soon as possible.

yaziine:
This PR fixes #99.

The fix is really trivial, just checking that the index we are iterating on is associated to the table we want to drop.

I take the opportunity to improve tests for `sql/query/drop.go` and I also added a method for listing the indexes (it's a good start for #100).

### DIFF
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -124,7 +124,7 @@ func (tx Transaction) DropTable(name string) error {
 			return err
 		}
 
-		// Remove only indexes associated to the target table.
+		// Remove only indexes associated with the target table.
 		if opts.TableName != name {
 			continue
 		}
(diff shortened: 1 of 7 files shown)


database/transaction.go
```diff
@@ -124,6 +124,11 @@ func (tx Transaction) DropTable(name string) error {
 			return err
 		}
 
+		// Remove only indexes associated to the target table.
```
asdine, line 127:
```suggestion
		// Remove only indexes associated with the target table.
```

PR #102, opened 2020-06-24
Add EXPLAIN statement

asdine:
This PR adds basic support for the EXPLAIN statement. If the query is a SELECT, UPDATE or DELETE, It displays the query plan tree, otherwise, it returns an error.
This PR also contains some fixes on how values are marshaled into JSON


document/json.go
```diff
@@ -117,12 +121,34 @@ func (vb *ValueBuffer) UnmarshalJSON(data []byte) error {
...
+	if err != nil {
+		return err
+	}
+
+	_, err = w.Write(buf)
+	if err != nil {
+		return err
+	}
+
+	_, err = w.Write([]byte{'\n'})
```
yaziine, line 134: Why do we need to add a `\n` at the end of the buffer? Same for `ArrayToJSON` function.

sql/parser/explain.go
```diff
@@ -0,0 +1,25 @@
+package parser
+
+import (
+	"github.com/genjidb/genji/sql/planner"
+	"github.com/genjidb/genji/sql/query"
+	"github.com/genjidb/genji/sql/scanner"
+)
+
+// parseDropStatement parses a drop string and returns a Statement AST object.
+// This function assumes the DROP token has already been consumed.
```
yaziine, line 10: Don't forget to update this.

sql/parser/explain.go
```diff
@@ -0,0 +1,25 @@
...
+func (p *Parser) parseExplainStatement() (query.Statement, error) {
+	// ensure we don't have multiple EXPLAIN keywords
+	tok, pos, lit := p.ScanIgnoreWhitespace()
+	if tok == scanner.EXPLAIN {
+		return nil, newParseError(scanner.Tokstr(tok, lit), []string{"TABLE", "INDEX"}, pos)
```
yaziine, line 15:
```suggestion
		return nil, newParseError(scanner.Tokstr(tok, lit), []string{"SELECT", "UPDATE", "DELETE"}, pos)
```

sql/planner/operation_string.go
```diff
@@ -0,0 +1,33 @@
+// Code generated by "stringer -type=Operation"; DO NOT EDIT.
```
yaziine, line 1: What's the purpose of this file?

sql/query/expr/literal.go
```diff
@@ -93,6 +99,22 @@ func (l LiteralExprList) IsEqual(other Expr) bool {
 	return true
 }
 
+// String implements the fmt.Stringer interface.
```
yaziine, line 102: Some `String` methods miss this comment. We should add it to each exported String methods.

sql/planner/operation_string.go
asdine: It was generated by this tool https://godoc.org/golang.org/x/tools/cmd/stringer, it's a way to generate `String() string` methods to enums

sql/parser/explain.go
asdine: Good catch 👍

document/json.go
asdine: Great question, I'll add a comment

asdine: I have removed it because it didn't really make sense anymore. We'd better have this function return a json object without newline, and rely on `json.Encoder` if we want to control the behaviour.

PR #109, opened 2020-07-05
Safer default document key

### LINKED ISSUE #74: Primary keys may overlap with concurrent inserts

asdine:
This PR changes the way the default document key is generated on tables without primary keys. Previously, multiple transactions executed at the same time could generate the same default keys using the Badger engine, only one transaction would work and the other ones would fail with a `Transaction Conflict`.

This PR makes some major changes:
- it introduces the concept of `docid`, the equivalent of SQLite's `rowid`. Whenever a document is inserted in a table that doesn't contain a primary key, a docid is generated (that's already the current behavior, it's just that it has a new name now). All mentions to "docid" refer specifically to that generated default key, not the user-chosen primary key.
- the next docid is determined by getting the largest docid of the table, then incremented
- the latest docid is cached in the database instance
- the largest docid is no longer stored in the database config
- if the generated docid is greater than the maximum int64, we look for the smallest available docid in the table.

This PR also fixes a bug with the Badger engine iterator that used to not return the right key when reverse is true.
 
Fixes #74

### DIFF
--- a/database/table.go
+++ b/database/table.go
@@ -84,13 +84,22 @@ func (t *Table) GetDocument(key []byte) (document.Document, error) {
 	return &d, err
 }
 
+// generate a key for d based on the table configuration.
+// if the table has a primary key, it extracts the field from
+// the document, converts it to the targeted type and returns
+// its encoded version.
+// if there are no primary key in the table, a default
+// key is generated, called the document id.
+// this function looks up for the highest key in the table,
+// increments it, caches it in tableDocIDs and returns its encoded version.
+// Generating a default key is safe for concurrent access across
+// multiple transactions.
 func (t *Table) generateKey(d document.Document) ([]byte, error) {
 	cfg, err := t.cfgStore.Get(t.name)
 	if err != nil {
 		return nil, err
 	}
 
-	var key []byte
 	if pk := cfg.GetPrimaryKey(); pk != nil {
 		v, err := pk.Path.GetValue(d)
 		if err == document.ErrFieldNotFound {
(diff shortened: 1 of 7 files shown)


database/table.go
```diff
@@ -103,22 +109,79 @@ func (t *Table) generateKey(d document.Document) ([]byte, error) {
...
+	var i int64 = 1
+
+	for it.Seek(nil); it.Valid(); it.Next() {
+		if !bytes.Equal(it.Item().Key(), encoding.EncodeInt64(i)) {
+			return i, nil
+		}
+		i++
+	}
+
+	return 0, errors.New("reached maximum number of documents in a table")
```
yaziine, line 184: We can reach that point if `i == math.MaxInt64`?

engine/badgerengine/store.go
```diff
@@ -16,10 +16,17 @@ type Store struct {
 	name     string
 }
 
+// build a long key for each key of a store
+// in the form: storePrefix + <sep> + 0 + key.
+// the 0 is used to separate the actual key
+// from the rest of the prexix and to ensure
+// we can quickly access the latest key of the store
+// by replacing 0 by anything bigger.
```
yaziine, line 24: I am not sure to understand the last part, could you elaborate please?

database/table.go
asdine: If ` == math.MaxInt64 + 1` yes

engine/badgerengine/store.go
asdine: Some context: The Badger engine stores data from all the stores in the same namespace. There is no concept of separate store/bucket in Badger, so we need to namespace them ourselves, using prefixes.

The role function of this function is to generate a badger key for each key of our stores:

```
// in Genji
...
store.Put([]byte("a"), someDocument)

// is equivalent to this in Badger
k := buildKey(somePrefix, []byte("a"))
```

This function generates a key in a way that makes it simple for us the iterator to iterate over the keys of a particular store.
Long story short, whenever we want to use an Iterator to jump to the **last** key of the store, we want to do this in one step, without having to iterate first over all the keys of the store.

To do this, we use a trick. Here is what we need to do without that trick:
```go
const separator = "|"

// this function is a simplified version of how it CURRENTLY works
func buildDumbKey(storeName string, key string) string {
  return storeName + separator + key
}

buildDumbKey("foo", "bar") // foo|bar
buildDumbKey("foo", "aar") // foo|aar
buildDumbKey("foo", "b") // foo|b
buildDumbKey("coo", "bar") // coo|bar

// the badger keys now looks like this
coo|bar
foo|aar
foo|b
foo|bar
/*

// how do we get the last key of store foo?
// Currently, we try a trick but its bugged, without this trick we'd have to do this:

// create a NON reverse cursor
it := newCursor()

// iterate until the end of the prefix foo
for it.Seek("foo"); i.ValidForPrefix("foo"); it.Next() {
}
// we reached the end, we need to go back once to place the cursor on "foo|bar"
last := it.Prev()
*/

Now here is a simplified version of that trick
```go
// this function is a simplified version of how this PR works
func buildSmartKey(storeName string, key string) string {
 // note that '0' is the byte 0 in this pseudo code, not the character 0
  return storeName + separator + byte(0) + key
}

buildSmartKey("foo", "bar") // foo|0bar
buildSmartKey("foo", "aar") // foo|0aar
buildSmartKey("foo", "b") // foo|0b
buildSmartKey("coo", "bar") // coo|0bar

// to jump to the end

// create a reverse cursor this time
it := newCursor(reverse: true)

// this will seek for foo" + byte(1), which is higher than ANY key in the foo store, because 1 is bigger than 0
// I chose 1 here, but it could be any byte greater than 0 (which means 0 < theChosenByte < 256).
// We know that this key doesn't exists because no buildSmartKey uses 0 all the time, but thanks to this the
// cursor places us at the last key anyway. Why? Because when Seek doesn't find the key you're looking for, it moves to the next one, following the direction of the iterator.
it.Seek("foo" + byte(1))
```

Issue #95, opened 2020-06-12
Add ALTER TABLE ... RENAME TO ...


yaziine:
This is a sub-issue of #94 that focuses on renaming an existing table.

We should be able to do the following:
```
genji> CREATE TABLE foo;
genji> INSERT INTO foo VALUES {name: "John Doe"};
genji> ALTER TABLE foo RENAME TO bar;
genji> SELECT * FROM bar;
{
  "name": "John Doe"
}
genji> SELECT * FROM foo;
table not found
```
If the table `bar` already exists, we should return an error.

yaziine:
After reflexion I think that we need to change how the db is structured to handle the renaming of a table.

Currently, how the DB is structured forces us to duplicate all documents of a given table if we want to rename it.
Time complexity: O(n) and Space complexity O(n) as well.

What I suggest is to break the actual DB schema to save some complexity if we want to rename a table.

```
_genji.master:
       tables:
               foo: {hash: 123abc, config: 456def, indexes: [123, 456, 789]},
               bar: {hash: 111aaa, config: 789ghi, indexes: [aa1]}
123abc: <table's content>
111aaa: <table's content>
```

That way we can constantly rename a table (O(1)) because we only need to change the name of the table in `tables` section. I also put the indexes there to avoid looking through all indexes when looking for indexes of a given table.
The `hash` allows us to have a reference to the content of the table. I don't know yet how this hash can be computed though.

PR #110, opened 2020-07-06
Name table stores with unique identifier to prevent having to rename them

yaziine:
This is a first iteration for solving #95. The schema of the DB has been changed, a new way of referencing tables has been introduced. 

Before, the store `_genji.tables` contained the configuration of each table and each store were keyed by the name of the table. But now, `_genji.tables` still holds the configuration of the table but also a `storeID` that contains a generated random value that will be the key of the store which contains the content of the table. This change will allow us to easily rename a table without duplicating its content (see https://github.com/genjidb/genji/issues/95#issuecomment-652399206 for more details).

Major changes:
- `storeID` introduced
- functions of the `Transaction` interface receive a `[]byte` now
- `ListStores` has been removed from the `Transaction` interface
- package `engineutil` has been removed, I removed because the implementation wasn't up to date anymore, we can add it later if needed


database/config.go
```diff
@@ -134,46 +139,112 @@ func (f *FieldConstraint) ScanDocument(d document.Document) error {
...
+	ti.cfg = &TableConfig{}
+	return ti.cfg.ScanDocument(doc)
 }
 
-func (t *tableConfigStore) Replace(tableName string, cfg *TableConfig) error {
+type tableInfoStore struct {
+	st engine.Store
+}
+
+// Insert inserts a new tableInfo for the table tableName.
```
asdine, line 188:
```suggestion
// Insert a new tableInfo for the given table name.
```

database/config.go
```diff
@@ -134,46 +139,112 @@ func (f *FieldConstraint) ScanDocument(d document.Document) error {
...
+// If the generated storeID already exists, a new one will be generated
```
asdine, line 189:
```suggestion
// It automatically generates a unique storeID for that table.
```

database/config.go
```diff
@@ -134,46 +139,112 @@ func (f *FieldConstraint) ScanDocument(d document.Document) error {
...
+// until it doesn't already exist.
```
asdine, line 190:
```suggestion
```

database/config.go
```diff
@@ -134,46 +139,112 @@ func (f *FieldConstraint) ScanDocument(d document.Document) error {
...
 
-type tableConfigStore struct {
-	st engine.Store
+type tableInfo struct {
+	// storeID is a generated ID that acts as a key to reference this table.
+	// The first-4 bytes represents the timestamp in second and the last-2 bytes are
+	// randomly generated.
+	storeID [6]byte
+
+	cfg *TableConfig
```
asdine, line 148: Why do we need a separate `TableConfig` type? What about just keeping everything in the same structure?

```go
type TableInfo struct {
  StoreID [6]byte
  FieldConstraints []FieldConstraint
  LastKey int64
}
```

database/config.go
```diff
@@ -202,6 +272,28 @@ func (t *tableConfigStore) Delete(tableName string) error {
 	return err
 }
 
+func (t *tableInfoStore) ListTables() ([]string, error) {
+	it := t.st.NewIterator(engine.IteratorConfig{Reverse: false})
```
asdine, line 276: 🌷 It's reverse by default
```suggestion
	it := t.st.NewIterator(engine.IteratorConfig{})
```

database/table.go
```diff
@@ -421,14 +422,14 @@ func (t *Table) Truncate() error {
 	return t.Store.Truncate()
 }
 
-// TableName returns the name of the table.
-func (t *Table) TableName() string {
+// Name returns the name of the table.
```
asdine, line 425: 👍

database/transaction.go
```diff
@@ -12,19 +11,20 @@ import (
...
 type Transaction struct {
-	db         *Database
-	Tx         engine.Transaction
-	writable   bool
-	tcfgStore  *tableConfigStore
+	db       *Database
+	Tx       engine.Transaction
+	writable bool
+
+	infoStore  *tableInfoStore
```
asdine, line 27: We should be explicit here

```suggestion
	tableInfoStore  *tableInfoStore
```

index/index.go
```diff
@@ -97,14 +97,19 @@ func NewUniqueIndex(tx engine.Transaction, idxName string) *UniqueIndex {
...
-	var b strings.Builder
-	b.WriteString(StorePrefix)
-	b.WriteString(name)
-	b.WriteByte(separator)
-	b.WriteByte(byte(t))
+func buildIndexName(name []byte, t Type) []byte {
+	var buf bytes.Buffer
 
-	return b.String()
+	// We can deterministically set the size of the buffer.
```
asdine, line 103: 👏

database/config.go
yaziine: This is a good point. There is no strict reason behind this. I'll do that 👍

yaziine: After reflexion, this change will imply a loooot of diff. I suggest to merge this one, and I'll do that later on in a dedicated PR.

PR #113, opened 2020-07-09
UPDATE SET... Allow dot notation. 

tzzed:
clean repository.
Correction of #84


tzzed:
added test using test table, comparing global json instead of each value .

PR #114, opened 2020-07-09
Add ALTER TABLE ... RENAME TO feature

### LINKED ISSUE #95: Add ALTER TABLE ... RENAME TO ...

yaziine:
This PR fixes #95. 

We can now rename a table and all indexes linked to the table are updated as well.

### DIFF
--- a/database/config.go
+++ b/database/config.go
@@ -148,50 +148,32 @@ func newTableInfoStore(tx engine.Transaction) (*tableInfoStore, error) {
 }
 
 // Insert a new tableInfo for the given table name.
-// It automatically generates a unique storeID for that table.
-func (t *tableInfoStore) Insert(tx engine.Transaction, tableName string, info *TableInfo) ([]byte, error) {
+func (t *tableInfoStore) Insert(tx engine.Transaction, tableName string, info *TableInfo) error {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 
 	_, ok := t.tableInfos[tableName]
 	if ok {
-		return nil, ErrTableAlreadyExists
+		return ErrTableAlreadyExists
 	}
-
-	var found bool = true
-	var id [6]byte
-	for found {
-		id = generateStoreID()
-
-		found = false
-		for _, ti := range t.tableInfos {
-			if ti.storeID == id {
-				// A store with this id already exists.
-				// Let's generate a new one.
-				found = true
-				break
-			}
-		}
-	}
-	info.storeID = id
 
 	v, err := msgpack.EncodeDocument(info.ToDocument())
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	st, err := tx.GetStore([]byte(tableInfoStoreName))
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	err = st.Put([]byte(tableName), v)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	t.tableInfos[tableName] = *info
-	return info.storeID[:], err
+	return nil
 }
 
 func (t *tableInfoStore) Get(tableName string) (*TableInfo, error) {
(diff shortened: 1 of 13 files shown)


database/config.go
```diff
@@ -158,20 +158,23 @@ func (t *tableInfoStore) Insert(tableName string, info *TableInfo) ([]byte, erro
...
 	var id [6]byte
-	for {
-		id = generateStoreID()
-		_, err = t.st.Get(id[:])
-		if err == nil {
-			// A store with this id already exists.
-			// Let's generate a new one.
-			continue
+	// Generate a new storeID if the one in the info is empty.
+	if info.storeID == id {
```
asdine, line 162: To avoid this logic and since it's only useful the first time we create a table, maybe we should generate the store id outside of this function, in the `Insert` method.

database/config.go
```diff
@@ -345,6 +348,17 @@ func (t *indexStore) Get(indexName string) (*IndexConfig, error) {
 	return &idxopts, nil
 }
 
+func (t *indexStore) Update(cfg IndexConfig) error {
```
asdine, line 351: I think `Replace` would be a better name because we are literally replacing the IndexConfig by another one. Also, it's consistent with TableInfoStore#Replace

database/table.go
```diff
@@ -21,6 +21,11 @@ type Table struct {
 	infoStore *tableInfoStore
 }
 
+// Name returns the name of the table.
```
asdine, line 24: I suppose this is a reordering of the file? Is there any other change? If so we can keep it as is, otherwise please do the reordering in a separate PR otherwise it gets extremely difficult to review.

database/transaction.go
```diff
@@ -103,6 +103,39 @@ func (tx Transaction) GetTable(name string) (*Table, error) {
 	}, nil
 }
 
+// RenameTable renames a table named old to new.
+// If it doesn't exist, returns ErrTableNotFound.
+func (tx Transaction) RenameTable(old, new string) error {
```
asdine, line 108: We should avoid shadowing keywords, I'm surprised the linter didn't yell at you. Perhaps something like `oldName`, `newName` ?

database/transaction.go
```diff
@@ -103,6 +103,39 @@ func (tx Transaction) GetTable(name string) (*Table, error) {
 	}, nil
 }
 
+// RenameTable renames a table named old to new.
```
asdine, line 106: Renaming is always from something to something else.
```suggestion
// RenameTable renames a table.
```

database/transaction.go
```diff
@@ -103,6 +103,39 @@ func (tx Transaction) GetTable(name string) (*Table, error) {
...
+// If it doesn't exist, returns ErrTableNotFound.
```
asdine, line 107:
```suggestion
// If it doesn't exist, it returns ErrTableNotFound.
```

database/transaction.go
```diff
@@ -170,6 +203,16 @@ func (tx Transaction) CreateIndex(opts IndexConfig) error {
 	return tx.indexStore.Insert(opts)
 }
 
+// UpdateIndex updates the index cfg referenced by cfg.IndexName.
+func (tx Transaction) UpdateIndex(cfg IndexConfig) error {
```
asdine, line 207: I don't think this should be exported, it seems really specific to the renaming feature and can be dangerous in other contexts. There doesn't seem to be a reason to replace an index config other than for renaming its related table.

database/table.go
yaziine: It's only a reordering. I did that in a separate commit for precisely ease the review.

database/transaction.go
yaziine: The linter didn't tell me anything regarding this. Does your tell you something?

database/transaction.go
yaziine: Yeah very good point!

database/transaction.go
asdine: It doesn't, that's weird. Though it's a bad habit shadowing keywords, even if that sounds good in that context, plus being explicit in a function signature doesn't hurt

database/table.go
asdine: Cool, thanks

database/config.go
yaziine: Done

database/config.go
```diff
@@ -414,6 +381,17 @@ func (t *indexStore) Get(indexName string) (*IndexConfig, error) {
 	return &idxopts, nil
 }
 
+func (t *indexStore) Replace(cfg IndexConfig) error {
```
asdine, line 384: When replacing something, the API must explicitly state what is the key used to identify the thing we are trying to replace, even if that information is already contained in the value of the second parameter.

```suggestion
func (t *indexStore) Replace(idxName string, cfg IndexConfig) error {
```

database/transaction.go
```diff
@@ -66,23 +69,53 @@ func (tx *Transaction) Promote() error {
...
 	}
-	sid, err := tx.tableInfoStore.Insert(tx.Tx, name, info)
+
+	var found bool = true
+	var id [6]byte
+	for found {
+		id = generateStoreID()
+
+		found = false
+		for _, ti := range tx.tableInfoStore.tableInfos {
```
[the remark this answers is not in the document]
yaziine: Good catch! 👍

Issue #106, opened 2020-06-28
Ignore input with whitespace only


asdine:
Whenever we press return with no input or whitespace only input in the CLI, it invites the user to add a `;`.
We should simply ignore the input.

Current behavior:

```
genji> <press return, with or without whitespace>
... <press ';', then return>
genji>
```

Expected behavior:

```
genji> <press return, with or without whitespace>
genji>
```

yaziine:
If we run a query with whitespace after the `;`, it also invites to add a `;`.
```
genji> create table foo; <press return with whitespace after the ; >
... <press ';', then return>
genji>
```

PR #115, opened 2020-07-09
ignore whitespace. press enter with no input or input ended by whitespace 

tzzed:
this a fix of #106

```
genji> 
```

and 

```
genji> create table foo; <press return with whitespace after the ; >
... <press ';', then return>
genji>
```

both are fixed.


cmd/genji/shell/shell.go
```diff
@@ -192,12 +202,13 @@ func (sh *Shell) executeInput(in string) error {
 		return sh.runCommand(in)
 	// If it ends with a ";" we can run a query
 	case strings.HasSuffix(in, ";"):
```
asdine, line 204: I think we just need to ignore the query if it matches the condition we discussed in the issue.
Basically, I think all we need to do is this:
- trim the `in` variable to remove all withespaces
- ignore the command if `len(in) == 0`

cmd/genji/shell/shell.go
```diff
@@ -192,12 +202,13 @@ func (sh *Shell) executeInput(in string) error {
...
-		sh.multiLine = false
-		sh.livePrefix = in
-		err := sh.runQuery(sh.query)
-		sh.query = ""
-		return err
+		return sh.QueryError(in)
+	//If we press enter without any query
+	case len(in) == 0:
+		return sh.QueryError(in)
+	// If it ends by espace "; "
```
asdine, line 209: This check is not necessary if we trim `in`

cmd/genji/shell/shell.go
```diff
@@ -184,6 +184,16 @@ func (sh *Shell) execute(in string) {
 	}
 }
 
+// Query Error return an error if we press return with no input, or ended by whitespace
+func (sh *Shell) QueryError(in string) error {
```
asdine, line 188: The name of this function doesn't seem to have anything to do with the content of the function, does it?

cmd/genji/shell/shell.go
tzzed: Indeed.

tzzed:
implementation of @asdine's suggestion for a simpler version.

asdine:
@tzzed I can't merge your PR, you need to [rebase](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) it

PR #122, opened 2020-07-13
Fix input with whitespace or no input. issue #106

tzzed:
This a fix of #106.

the expected behavior is respected:

```
genji>
```

and

```
genji> create table foo; <press return with whitespace after the ; >
genji>
```


.gitignore
```diff
@@ -1,24 +0,0 @@
-# Binaries for programs and plugins
```
asdine, line 1: Is there a reason why you delete the gitignore?

Issue #105, opened 2020-06-28
Arithmetic operators should return null for incompatible types


asdine:
```sql
genji> insert into foo (a) values (1), ('hello'), (true);
genji> select * from foo where a + 1 > 0;
{
  "a": 1
}
type "text" incompatible with "integer"
```

```sql
# currently
genji> SELECT 1 + 'hello';
type "text" incompatible with "integer"

# expected
genji> SELECT 1 + 'hello';
{
  "1 + 'hello'": null
}
```

PR #123, opened 2020-07-13
fix issue #105  arithmetic operator between int and text value 

tzzed:
This is a fix of #105 

```
genji> select 1 + 'hello';
{
  "1 + 'hello'": null
}
genji>
```


tzzed:
test for all operators are sets

tzzed:
wrote tests for all operators

PR #125, opened 2020-07-15
Replace TableConfig by TableInfo

yaziine:
As suggested in https://github.com/genjidb/genji/pull/110#discussion_r451303589, this PR removes the `TableConfig` type and replaces it by the `TableInfo` one.

`TableInfo` now holds _all_ informations regarding a table.

### DIFF
--- a/database/table.go
+++ b/database/table.go
@@ -90,7 +90,7 @@ func (t *Table) GetDocument(key []byte) (document.Document, error) {
 	return &d, err
 }
 
-// generate a key for d based on the table constraints.
+// generate a key for d based on the table configuration.
 // if the table has a primary key, it extracts the field from
 // the document, converts it to the targeted type and returns
 // its encoded version.
--- a/database/table.go
+++ b/database/table.go
@@ -21,14 +21,14 @@ type Table struct {
 	infoStore *tableInfoStore
 }
 
-// Config of the table.
-func (t *Table) Config() (*TableConfig, error) {
+// Info of the table.
+func (t *Table) Info() (*TableInfo, error) {
 	ti, err := t.infoStore.Get(t.name)
 	if err != nil {
 		return nil, err
 	}
 
-	return &TableConfig{FieldConstraints: ti.FieldConstraints}, nil
+	return ti, nil
 }
 
 type encodedDocumentWithKey struct {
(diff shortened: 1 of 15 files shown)


database/table.go
```diff
@@ -89,7 +90,7 @@ func (t *Table) GetDocument(key []byte) (document.Document, error) {
 	return &d, err
 }
 
-// generate a key for d based on the table configuration.
+// generate a key for d based on the table constraints.
```
asdine, line 93: The table configuration is a generic term that means "based on how the table is configured". "table constraints" however is too specific, as it refers to SQL Table Constraints, which we don't support yet.
Is there a reason you want to change this line?

sql/parser/create.go
```diff
@@ -46,8 +46,8 @@ func (p *Parser) parseCreateTableStatement() (query.CreateTableStmt, error) {
 		return stmt, err
 	}
 
-	// parse table config
-	err = p.parseTableConfig(&stmt.Config)
+	// parse fields constraints
+	err = p.parseFieldsConstraints(&stmt.Info)
```
asdine, line 50:
```suggestion
	// parse field constraints
	err = p.parseFieldConstraints(&stmt.Info)
```

sql/parser/create.go
```diff
@@ -75,7 +75,7 @@ func (p *Parser) parseIfNotExists() (bool, error) {
 	return true, nil
 }
 
-func (p *Parser) parseTableConfig(cfg *database.TableConfig) error {
+func (p *Parser) parseFieldsConstraints(info *database.TableInfo) error {
```
asdine, line 78:
```suggestion
func (p *Parser) parseFieldConstraints(info *database.TableInfo) error {
```

database/table.go
yaziine: I agree, it's indeed too specific. I'll change it like it was before. 👍

sql/parser/create.go
yaziine: I don't agree because this method parses _all_ fields constraints and not only one. Wdyt?

asdine: "Field constraint" is a compound word. When you write compound words in plural form, only the last one takes an `s`.

Example: one train station, two train stations, one car park, two car parks

📖 https://www.myenglishpages.com/site_php_files/grammar-lesson-plural-compound-nouns.php

PR #126, opened 2020-07-15
Ignore Input shell with whitespaces only or no input

### LINKED ISSUE #106: Ignore input with whitespace only

tzzed:
This PR fixes #106 by adding `TrimSpace` of input.
The command is ignored if the input is empty.

### DIFF
--- a/cmd/genji/shell/shell.go
+++ b/cmd/genji/shell/shell.go
@@ -185,6 +185,7 @@ func (sh *Shell) execute(in string) {
 }
 
 func (sh *Shell) executeInput(in string) error {
+	in = strings.TrimSpace(in)
 	switch {
 	// if it starts with a "." it's a command
 	// it must not be in the middle of a multi line query though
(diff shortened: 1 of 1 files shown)


cmd/genji/go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/mattn/go-runewidth v0.0.7 // indirect
 	github.com/mattn/go-tty v0.0.3 // indirect
 	github.com/pkg/term v0.0.0-20190109203006-aa71e9d9e942 // indirect
+	github.com/stretchr/testify v1.5.1
```
yaziine, line 12: I think this dependency was mistakenly added.

cmd/genji/shell/shell.go
```diff
@@ -198,6 +199,9 @@ func (sh *Shell) executeInput(in string) error {
 		err := sh.runQuery(sh.query)
 		sh.query = ""
 		return err
+	// If press enter with no input
```
yaziine, line 202: 🌷  I suggest this instead, what do you think?
```suggestion
	// If the input is empty we ignore it
```

yaziine:
Thanks for the contribution @tzzed, I added some suggestion/comment. Feel free to ask questions if something is not clear!

tzzed:
Dependency deleted

PR #127, opened 2020-07-15
Allow dot notation for update set.

### LINKED ISSUE #84: Support dot notation on UPDATE ... SET

tzzed:
This PR fixes #84 by a deep recursive course of the document based on the ValuePath.
 #Set method is now taking a ValuePath.

### DIFF
--- a/document/document.go
+++ b/document/document.go
@@ -86,8 +86,8 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
 	return Value{}, ErrFieldNotFound
 }
 
-// SetFieldValue replaces a field if it already exists or creates one if not.
-func (fb *FieldBuffer) SetFieldValue(field string, reqValue Value) error {
+// setFieldValue replaces a field if it already exists or creates one if not.
+func (fb *FieldBuffer) setFieldValue(field string, reqValue Value) error {
 	_, err := fb.GetByField(field)
 	switch err {
 	case ErrFieldNotFound:
@@ -101,9 +101,9 @@ func (fb *FieldBuffer) SetFieldValue(field string, reqValue Value) error {
 	return err
 }
 
-// SetValueFromValue deep replaces or creates a field
-// through the value path to get the value to set or replace.
-func (fb *FieldBuffer) SetValueFromValue(v Value, p ValuePath, newValue Value) (Value, error) {
+// setValueAtPath deep replaces or creates a field
+// through the value path to get the value and create or replace it.
+func (fb *FieldBuffer) setValueAtPath(v Value, p ValuePath, newValue Value) (Value, error) {
 
 	switch v.Type {
 	case DocumentValue:
@@ -114,7 +114,7 @@ func (fb *FieldBuffer) SetValueFromValue(v Value, p ValuePath, newValue Value) (
 		}
 
 		if len(p) == 1 {
-			err = buf.SetFieldValue(p[0], newValue)
+			err = buf.setFieldValue(p[0], newValue)
 			return NewDocumentValue(&buf), err
 		}
 
@@ -123,12 +123,12 @@ func (fb *FieldBuffer) SetValueFromValue(v Value, p ValuePath, newValue Value) (
 			return v, err
 		}
 
-		va, err = buf.SetValueFromValue(va, p[1:], newValue)
+		va, err = buf.setValueAtPath(va, p[1:], newValue)
 		if err != nil {
 			return v, err
 		}
 
-		err = buf.SetFieldValue(p[0], va)
+		err = buf.setFieldValue(p[0], va)
 		return NewDocumentValue(&buf), err
 	case ArrayValue:
 		var vb ValueBuffer
@@ -136,7 +136,6 @@ func (fb *FieldBuffer) SetValueFromValue(v Value, p ValuePath, newValue Value) (
 		if err != nil {
 			return v, err
 		}
-
 
 		index, err := strconv.Atoi(p[0])
 		if err != nil {
@@ -153,7 +152,7 @@ func (fb *FieldBuffer) SetValueFromValue(v Value, p ValuePath, newValue Value) (
 			return NewArrayValue(&vb), err
 		}
 
-		va, err = fb.SetValueFromValue(va, p[1:], newValue)
+		va, err = fb.setValueAtPath(va, p[1:], newValue)
 		err = vb.Replace(index, va)
 		return NewArrayValue(&vb), err
 	}
@@ -166,7 +165,7 @@ func (fb *FieldBuffer) Set(path ValuePath, reqValue Value) error {
 	// check if the ValuePath contains only one field to set or replace
 	if len(path) == 1 {
 		// Set or replace the unique field
-		return fb.SetFieldValue(path[0], reqValue)
+		return fb.setFieldValue(path[0], reqValue)
 	}
 
 	for i, field := range fb.fields {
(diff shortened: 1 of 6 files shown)


document/document.go
```diff
@@ -86,16 +86,108 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
 	return Value{}, ErrFieldNotFound
 }
 
+// SetFieldValue replaces a field if it already exists or creates one if not.
+func (fb *FieldBuffer) SetFieldValue(field string, reqValue Value) error {
```
asdine, line 90: I don't think it's worth it to export this method. Also, this only basically does an if and an else, I don't think it deserves being a function.

document/document.go
```diff
@@ -86,16 +86,108 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+		_ = fb.Replace(field, reqValue)
+		return nil
+	}
+
+	return err
+}
+
+// SetValueFromValue deep replaces or creates a field
+// through the value path to get the value to set or replace.
+func (fb *FieldBuffer) SetValueFromValue(v Value, p ValuePath, newValue Value) (Value, error) {
```
asdine, line 106: Whenever a method starts with an uppercase letter, it means that it's exported and can be used by someone else's package. It is really important to choose carefully what we want to expose.
For example here, based on the explanation, what is the difference between this method and `Set`?
As a user of this package, I only need to be able to set a value at a given path, and `Set` seems like the right choice for that, why would I need to use `SetValueFromValue`?

This method should:
- be unexported: this logic is internal to this package and shouldn't be exposed.
- be named correctly: what exactly are you trying to do with it? Something like `setValueAtPath` that would be better

document/document.go
tzzed: Ooh ! The parser is a regression, I don't know what I did bad. So I going to correct it.

document/document.go
tzzed: I am agree about the naming `setValueAtPath ` will be better and not exported effectively.

document/document.go
```diff
@@ -86,16 +86,107 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+}
+
+// setValueAtPath deep replaces or creates a field
+// through the value path to get the value and create or replace it.
+func (fb *FieldBuffer) setValueAtPath(v Value, p ValuePath, newValue Value) (Value, error) {
+
+	switch v.Type {
+	case DocumentValue:
+		var buf FieldBuffer
+		err := buf.Copy(v.V.(Document))
```
asdine, line 111: On line 174, we are already copying the entire Document, I don't think we need to do it here right?
I think replacing l174 and this line by `buf.ScanDocument` instead will be more efficient as it won't decode values that are untouched

document/document.go
```diff
@@ -86,16 +86,107 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+func (fb *FieldBuffer) Set(path ValuePath, reqValue Value) error {
+	// check if the ValuePath contains only one field to set or replace
+	if len(path) == 1 {
+		// Set or replace the unique field
+		return fb.setFieldValue(path[0], reqValue)
+	}
+
+	for i, field := range fb.fields {
+		if path[0] == field.Field {
+			var buf FieldBuffer
```
asdine, line 173: Why do we need to deep copy `fb`, since it's already a FieldBuffer? Is there a reason why we don't modify it directly?

document/document.go
```diff
@@ -86,16 +86,107 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+	return v, nil
+}
+
 // Set replaces a field if it already exists or creates one if not.
-func (fb *FieldBuffer) Set(f string, v Value) {
-	for i := range fb.fields {
-		if fb.fields[i].Field == f {
+func (fb *FieldBuffer) Set(path ValuePath, reqValue Value) error {
+	// check if the ValuePath contains only one field to set or replace
+	if len(path) == 1 {
```
asdine, line 166: Do we need this special case? If I understand correctly, it seems that it is naturally handled by the loop below

document/document.go
```diff
@@ -86,16 +86,107 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+				return err
+			}
 			fb.fields[i].Value = v
-			return
+			return nil
 		}
 	}
 
-	fb.Add(f, v)
+	//return Err if the request is like foo.1.2.etc where foo doesn't exist
```
asdine, line 188:
```suggestion
	// return an error if the first element of the path is not found
```

sql/parser/update.go
```diff
@@ -56,12 +57,18 @@ func (p *Parser) parseSetClause() ([]updateSetPair, error) {
 			}
 		}
 
-		// Scan the identifier for the field name.
 		tok, pos, lit := p.ScanIgnoreWhitespace()
```
asdine, line 60: There is no need to scan for an identifier if we are going to parse a field ref, I think this should be removed

sql/parser/update.go
```diff
@@ -56,12 +57,18 @@ func (p *Parser) parseSetClause() ([]updateSetPair, error) {
...
 		if tok != scanner.IDENT {
 			return nil, newParseError(scanner.Tokstr(tok, lit), []string{"identifier"}, pos)
 		}
 
+		p.Unscan()
+		ref, err := p.parseFieldRef()
+		if err != nil {
+			return nil, newParseError(scanner.Tokstr(tok, lit), []string{"identifier"}, pos)
```
asdine, line 68: `parseFieldRef` already returns an error, we must return it
```suggestion
			return nil, err
```

document/document.go
```diff
@@ -86,16 +86,107 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+		return nil
+	}
+
+	return err
+}
+
+// setValueAtPath deep replaces or creates a field
+// through the value path to get the value and create or replace it.
+func (fb *FieldBuffer) setValueAtPath(v Value, p ValuePath, newValue Value) (Value, error) {
+
```
asdine, line 107:
```suggestion
```

document/document.go
```diff
@@ -86,16 +86,107 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+	case nil:
+		_ = fb.Replace(field, reqValue)
+		return nil
+	}
+
+	return err
+}
+
+// setValueAtPath deep replaces or creates a field
+// through the value path to get the value and create or replace it.
```
asdine, line 105:
```suggestion
// using the value path.
```

document/document.go
tzzed: You 're right

document/document.go
tzzed: I didn't know if the `ScanDocument` function was enough.

document/document.go
tzzed: It is  for the `ValuePath` with single value. In the loop we call `fb.setValueAtPath(field.Value, path[1:], reqValue)` and give it a slice that start to the next value of the `ValuePath`.

sql/parser/update.go
tzzed: Indeed

sql/parser/update.go
tzzed: one more, indeed!

asdine (approved):
Thank you @tzzed ! Could you rebase with master please?

PR #129, opened 2020-07-17
Command .indexes displaying indexes of database or a table.

### LINKED ISSUE #100: Add .indexes command

tzzed:
This PR fixes #100. Two options for the command `.Indexes` and `.Indexes <table name>` are implemented. Tests `.tables` and `.indexes` are added. 

```
genji> .indexes
main skill index on users(skills.0)
idx_nen on users(nen)
genji>
```

Refactoring of  `struct Index`. Let me know what you think about it

### DIFF
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -2,11 +2,16 @@ package shell
 
 import (
 	"fmt"
-
 	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/database"
+	"github.com/genjidb/genji/document"
 )
 
-func runTablesCmd(db *genji.DB) error {
+func runTablesCmd(db *genji.DB, cmd []string) error {
+	if len(cmd) > 1 {
+		return fmt.Errorf("too many arguments in call to %s", cmd[0])
+	}
+
 	var tables []string
 	err := db.View(func(tx *genji.Tx) error {
 		var err error
@@ -23,4 +28,59 @@ func runTablesCmd(db *genji.DB) error {
 	}
 
 	return nil
+}
+
+// runTableIndexCmd shows the all indexes that the given table contains.
+func runTableIndexCmd(db *genji.DB, tableName string) error {
+	err := db.ViewTable(tableName, func(tx *genji.Tx, table *database.Table) error {
+		return table.PrintIndexes()
+	})
+
+	if err == document.ErrFieldNotFound {
+		return nil
+	}
+
+	return err
+}
+
+// runAllIndexesCmd shows all indexes that the database contains.
+func runAllIndexesCmd(db *genji.DB) error {
+	var tables []string
+	err := db.View(func(tx *genji.Tx) error {
+		var err error
+		tables, err = tx.ListTables()
+		return err
+	})
+
+	if err != nil {
+		return err
+	}
+
+	for _, table := range tables {
+		// If there is no index in a table we continue to the next. No error handling needed
+		_ = db.View(func(tx *genji.Tx) error {
+			t, err := tx.GetTable(table)
+			if err != nil {
+				return err
+			}
+
+			return t.PrintIndexes()
+		})
+	}
+
+	return nil
+}
+
+// runIndexesCmd select a kind of indexes command is wanted
+func runIndexesCmd(db *genji.DB, in []string) error {
+	switch len(in) {
+	case 1:
+		// If the input is ".Indexes"
+		return runAllIndexesCmd(db)
+	case 2:
+		// If the input is ".Indexes <tableName>" cmd[1] is the table name
+		return runTableIndexCmd(db, in[1])
+	}
+
+	return fmt.Errorf("too many arguments in call to %s", in[0])
 }
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -2,14 +2,13 @@ package shell
 
 import (
 	"fmt"
+
 	"github.com/genjidb/genji"
-	"github.com/genjidb/genji/database"
-	"github.com/genjidb/genji/document"
 )
 
 func runTablesCmd(db *genji.DB, cmd []string) error {
 	if len(cmd) > 1 {
-		return fmt.Errorf("too many arguments in call to %s", cmd[0])
+		return fmt.Errorf("usage: .tables")
 	}
 
 	var tables []string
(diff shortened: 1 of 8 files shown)


cmd/genji/shell/shell.go
```diff
@@ -189,7 +189,7 @@ func (sh *Shell) executeInput(in string) error {
 	switch {
 	// if it starts with a "." it's a command
 	// it must not be in the middle of a multi line query though
-	case strings.HasPrefix(in, ".") && sh.query == "":
+	case strings.HasPrefix(in, "."):
```
yaziine, line 192: I am not sure to understand why did you remove the `&& sh.query == ""` statement. Could you elaborate please?

cmd/genji/shell/shell.go
```diff
@@ -213,16 +213,29 @@ func (sh *Shell) executeInput(in string) error {
 	return nil
 }
 
-func (sh *Shell) runCommand(cmd string) error {
-	switch cmd {
+func (sh *Shell) runCommand(in string) error {
+	cmd := strings.Fields(in)
```
yaziine, line 217: TIL! I used to use `strings.Split` for this kind of action. Thanks!

cmd/genji/shell/shell.go
```diff
@@ -213,16 +213,29 @@ func (sh *Shell) executeInput(in string) error {
...
 			return err
 		}
-		return runTablesCmd(db)
+
+		return runTablesCmd(db, cmd)
 	case ".exit":
+		if len(cmd) > 1 {
+			return fmt.Errorf("too many arguments in call to %s", cmd[0])
+		}
+
```
yaziine, line 230: I think that how it was done before was enough. Could you tell me what motivates you to change this? There is probably a use case that I don't see.

database/config.go
```diff
@@ -298,10 +298,7 @@ func (i *IndexConfig) ScanDocument(d document.Document) error {
 type Index struct {
 	index.Index
 
-	IndexName string
-	TableName string
-	Path      document.ValuePath
-	Unique    bool
+	Opts IndexConfig
```
yaziine, line 301: First of all thanks for doing this! 

However, I suggest to embed directly `IndexConfig`:
```go
type Index struct {
    index.Index
    IndexConfig
}
```
like that instead of accessing the config fields through `Opts`
```go
idx.Opts.Path.GetValue(d)
```
we can directly do
```go
idx.Path.GetValue(d)
````

What do you think?

File: database/table.go [hunk 5]
```diff
@@ -554,3 +551,17 @@ func (t *Table) Indexes() (map[string]Index, error) {
 
 	return indexes, nil
 }
+
+// PrintIndexes prints a list of indexes of the table receiver.
+func (t *Table) PrintIndexes() error {
```
yaziine, line 556: Printing the indexes is a feature that is really linked to the shell. Thus, I suggest to move this logic to its package. 
I think that it's useful to have this function only there. What do you think?

cmd/genji/shell/command_test.go
```diff
@@ -0,0 +1,82 @@
+package shell
+
+import (
+	"github.com/genjidb/genji"
+	"github.com/stretchr/testify/require"
+	"strings"
+	"testing"
+)
+
+func Test_runTablesCmd(t *testing.T) {
```
yaziine, line 10: I think we should have the same way of naming function tests within the codebase. Does it make sense?
```suggestion
func TestTablesCmd(t *testing.T) {
```

cmd/genji/shell/command_test.go
```diff
@@ -0,0 +1,82 @@
...
+			defer db.Close()
+
+			if err := runTableIndexCmd(db, test.in); (err != nil) != test.wantErr {
+				require.Errorf(t, err, "", test.wantErr)
+			}
+		})
+	}
+}
+
+func Test_runIndexesCmd(t *testing.T) {
```
yaziine, line 40: Same as above
```suggestion
func TestIndexesCmd(t *testing.T) {
```

cmd/genji/shell/command.go
```diff
@@ -24,3 +29,62 @@ func runTablesCmd(db *genji.DB) error {
...
+		//
+		if err == database.ErrTableNotFound {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// runIndexesCmd executes all indexes of the database or all indexes of the given table.
```
yaziine, line 78:
```suggestion
// runIndexesCmd executes the .indexes command.
```

cmd/genji/shell/command.go
```diff
@@ -24,3 +29,62 @@ func runTablesCmd(db *genji.DB) error {
...
+func runIndexesCmd(db *genji.DB, in []string) error {
+	switch len(in) {
+	case 1:
+		// If the input is ".Indexes"
```
yaziine, line 82:
```suggestion
		// If the input is ".indexes"
```

cmd/genji/shell/command.go
```diff
@@ -24,3 +29,62 @@ func runTablesCmd(db *genji.DB) error {
...
+		return runAllIndexesCmd(db)
+	case 2:
+		// If the input is ".Indexes <tableName>" cmd[1] is the table name
```
yaziine, line 85:
```suggestion
		// If the input is ".indexes <tableName>"
```

cmd/genji/shell/command.go
```diff
@@ -24,3 +29,62 @@ func runTablesCmd(db *genji.DB) error {
...
+		return runTableIndexCmd(db, in[1])
+	}
+
+	return fmt.Errorf("too many arguments in call to %s", in[0])
```
yaziine, line 89:
```suggestion
	return fmt.Errorf("too many arguments to the command %s", in[0])
```

database/config.go
yaziine: Actually, I am sure its a good idea because it will expose all method of the `IndexConfig`. What do you think @asdine ?

cmd/genji/shell/shell.go
tzzed: I did that because `.indexes` could take an argument `.indexes tableName` that led me to change the behaviour of`runCommand` by checking the len of the splitted input string. 
`sh.query == ""` was a 'guard' against input command with arguments but now we need it for `.indexes`.

cmd/genji/shell/shell.go
tzzed: See two comments above.

database/table.go
tzzed: I got it.

database/config.go
tzzed: Very good thing this avoids the refactoring of the majority of files.

asdine: Yeah I would keep it like this, embedding would make the `Index` type API weird

cmd/genji/shell/command.go
```diff
@@ -24,3 +29,62 @@ func runTablesCmd(db *genji.DB) error {
...
 	return nil
 }
+
+// runTableIndexCmd prints the all indexes that the given table contains.
+func runTableIndexCmd(db *genji.DB, tableName string) error {
+	err := db.ViewTable(tableName, func(tx *genji.Tx, table *database.Table) error {
+		return table.PrintIndexes()
+	})
+
+	if err == document.ErrFieldNotFound {
```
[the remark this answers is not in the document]
tzzed: I tried to catch it before but no success, I should catch it before and did not break the error handler for another command or request.Need more investigation...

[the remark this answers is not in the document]
asdine: If you need help don't hesitate asking 👍

cmd/genji/shell/command_test.go
```diff
@@ -0,0 +1,82 @@
+package shell
+
+import (
+	"github.com/genjidb/genji"
```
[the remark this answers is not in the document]
tzzed: now yes!

sql/query/insert.go
```diff
@@ -54,6 +55,7 @@ func (stmt InsertStmt) insertDocuments(t *database.Table, stack expr.EvalStack)
 	var res Result
 
 	for _, e := range stmt.Values {
+		fmt.Println("values row:\n", e)
```
asdine, line 58: You forgot to remove this debug line 😉

cmd/genji/shell/command.go
yaziine: @tzzed this comment has been marked as resolved but the suggestion is not applied. Probably due to an omission. Do you think you can apply it?

cmd/genji/shell/shell.go
```diff
@@ -213,16 +213,29 @@ func (sh *Shell) executeInput(in string) error {
...
 		db, err := sh.getDB()
 		if err != nil {
 			return err
 		}
-		return runTablesCmd(db)
+
+		return runTablesCmd(db, cmd)
 	case ".exit":
+		if len(cmd) > 1 {
+			return fmt.Errorf("too many arguments in call to %s", cmd[0])
```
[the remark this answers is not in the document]
yaziine: Same here @tzzed

PR #130, opened 2020-07-18
Simplifying the type system

asdine:
This PR simplifies the type system by:
- removing fixed int*
- keeping the `integer` type, whose encoding size depends on the size of value it holds
- renaming `float64` to `double` to fit SQL standards
- reassigning new values to `document.ValueType` enum so that adding new types in the future wouldn't require us to modify them

All of these changes can be seen in the `document/value.go` file, the rest of the PR is only about removing the use of the old types.


document/value.go
```diff
@@ -93,12 +93,12 @@ func (t ValueType) IsNumber() bool {
 
 // IsInteger returns true if t is a signed or unsigned integer of any size.
 func (t ValueType) IsInteger() bool {
-	return t >= Int8Value && t <= Int64Value || t == DurationValue
+	return t == IntegerValue || t == DurationValue
 }
 
-// IsFloat returns true if t is either a Float32 or Float64.
+// IsFloat returns true if t is a Double.
 func (t ValueType) IsFloat() bool {
```
yaziine, line 100: What about renaming this to `IsDouble`?

document/value.go
```diff
@@ -131,56 +122,52 @@ func NewBoolValue(x bool) Value {
...
-}
-
-// NewInt8Value encodes x and returns a value.
-func NewInt8Value(x int8) Value {
+func NewIntegerValue(x int64) Value {
 	return Value{
-		Type: Int8Value,
-		V:    x,
+		Type: IntegerValue,
+		V:    int64(x),
```
yaziine, line 130: The PR description says:
> whose encoding size depends on the size of value it holds

I don't see where this is done?

document/value.go
asdine: This is a good question that I was asking myself for a while. But it depends on what we want to answer here:
- Do we want to know if the type is from the `float` family (historically we had Float32 and Float64 support)
- or do we want to know if the type is a double?

The first question doesn't really need to be answered since we only have one float type now. 
The second question can be answered by doing `v.Type == document.DoubleValue`.

I think this function will get removed eventually along others of this package, but I think it deserves another PR

document/value.go
asdine: This is done in the MessagePack package, everything about encoding is in the `document/encoding` package

document/value.go
yaziine: > I think it deserves another PR

Definitely. Let's keep it as it is for now if it will be removed later on.

PR #131, opened 2020-07-19
Fix parentheses behaviour

asdine:
This PR changes how parentheses were parsed by the parser. Previously, they were considered as lists of expressions, which after evaluation turned into an array. This behaviour prevented us from using parentheses as arithmetic sub expressions.

Before:
```sql
genji> select ((10 + 5 + 5) * 4) / 2;
{
  "((10 + 5 + 5) * 4) / 1": null
}
genji> select ((10 + 5 + 5) * 4);
{
  "((10 + 5 + 5) * 4)": [
    null
  ]
}
```

After:
```sql
genji> select ((10 + 5 + 5) * 4) / 2;
{
  "((10 + 5 + 5) * 4) / 1": 40
}
genji> select ((10 + 5 + 5) * 4);
{
  "((10 + 5 + 5) * 4)": 80
}
```

### DIFF
--- a/sql/query/expr/expr.go
+++ b/sql/query/expr/expr.go
@@ -119,3 +119,16 @@ type Operator interface {
 	SetRightHandExpr(Expr)
 	Token() scanner.Token
 }
+
+// Parentheses is a special expression which turns
+// any expression as a unary expression.
+// It hides the underlying expression operator from the parser
+// so that it doesn't get reordered by precedence.
+type Parentheses struct {
+	E Expr
+}
+
+// Eval calls the underlying expression Eval method.
+func (p Parentheses) Eval(es EvalStack) (document.Value, error) {
+	return p.E.Eval(es)
+}
--- a/sql/query/expr/expr.go
+++ b/sql/query/expr/expr.go
@@ -121,8 +121,8 @@ type Operator interface {
 }
 
 // Parentheses is a special expression which turns
-// any expression as a unary expression.
-// It hides the underlying expression operator from the parser
+// any sub-expression as unary.
+// It hides the underlying operator, if any, from the parser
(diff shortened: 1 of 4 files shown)


sql/query/expr/expr.go
```diff
@@ -119,3 +119,16 @@ type Operator interface {
 	SetRightHandExpr(Expr)
 	Token() scanner.Token
 }
+
+// Parentheses is a special expression which turns
+// any sub-expression as unary.
+// It hides the underlying operator, if any, from the parser
+// so that it doesn't get reordered by precedence.
```
yaziine, line 126: What does that mean exactly?

asdine: In SQL, operators are ordered by precedence, for example, `*` has higher precedence than `+`, which means it must be evaluated first:

```python
3 + 4 * 2
# 4 * 2 is evaluated first
3 + 8
# then + is evaluated
11
```

If you want to avoid that, you can use parentheses

```python
(3 + 4) * 2
# 3 + 4 is evaluated first
7 * 2
# then * is evaluated
14
```

This behavior is done in the parser, when parsing the expression: https://github.com/genjidb/genji/blob/454cc983db979394c310a5a5df0e1180fb0f1447/sql/parser/expr.go#L30-L78

This function builds a tree based on operator precedence.
To avoid an expression in parentheses to have its precedence analyzed and taken into account by that function ([these two lines](https://github.com/genjidb/genji/blob/454cc983db979394c310a5a5df0e1180fb0f1447/sql/parser/expr.go#L70-L71)) we "hide" it by wrapping it in a type that **doesn't** implement the Operator interface so that in the eye of these two lines, that expression is unary, not binary.

PR #153, opened 2020-08-19
Add prefix to table stores

### LINKED ISSUE #73: Define and filter forbidden characters on identifiers
A few packages (`index`, `engine/*`, ...) use special characters as separators when they build keys (example: https://github.com/asdine/genji/blob/master/index/index.go#L98-L106).
Currently, we don't prevent users to use these separators in their documents, which can cause problems.
Define these characters somewhere and make sure to return an error if they are used

asdine:
This PR adds a prefix to every created table store. It also fixes a panic when decoding a tableInfo.
Fixes #73

### DIFF
--- a/database/config.go
+++ b/database/config.go
@@ -14,6 +14,8 @@ import (
 	"github.com/genjidb/genji/index"
 )
 
+const storePrefix = 't'
+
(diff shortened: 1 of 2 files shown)


yaziine (approved):
LGTM!

Issue #143, opened 2020-07-30
Add tables suggestion to the CLI


yaziine:
After the `FROM` keyword, the cli should suggest us all existing tables.
```
genji> SELECT * FROM <suggestion of all existing tables>
```

yaziine:
It would be nice if we add the suggestions on all statement that require the `FROM` clause. 
For example, in the `ALTER` and `UPDATE` ones.

PR #159, opened 2020-08-21
Table suggestion after FROM keyword

tzzed:
This PR fixes the #143 .
Tables are listed after `FROM` keyword in `SELECT and DELETE` statements.
Added to `DROP TABLE` statement too.
If there is no tables the word `table_name` is displayed:
```
genji> SELECT * FROM table_name
```


asdine (approved):
Thank you!

PR #161, opened 2020-08-22
New way of referencing a value

asdine:
Currently, we are using the dot-notation to refer to a particular value in a document. This notation is limited for several reasons:
- it doesn't feel natural: accessing array indexes using dots instead of square brackets feels weird
- it doesn't allow specifying ranges

This PR changes that by using a different notation, that is closer to what is used in Go, Python, Javascript, etc.
Here are some examples:

```go
// simple field 
a

// field with backquotes
`   some "field" `

// nested fields
a.b
a.` b `.c

// array indexes
a[0]
a.b[100][10].c
```

This will allow us to introduce interesting features in the future:

```go
// slicing
a[1:3]
a[1:]

// accessing last element
a[-1]
// accessing second last element
a[-2]

// expressions
a[1 + 3]
a[ARRAY_LENGTH(a) - 4]
```

I believe this notation feels more natural since it's the same as in mainstream programming languages


tzzed:
When the field using backquotes with spaces like this `"   some \"field\" "`, the spaces must be deleted or not?
input: ` \`"   some \"field\" \`` => output: `" some \"field\" ",` 
Is the expected behaviour?

asdine:
@tzzed Please open an issue if you see something odd, discussing in closed pull requests is not ideal

PR #164, opened 2020-08-25
Commands suggestions

### LINKED ISSUE #163: Add commands suggestions to the CLI
I think it would be useful to have a suggestion commands when we enter `.` in the shell.
```
genji>. <suggestion of all commands>
```

tzzed:
This PR fixes #163. It suggests command when we enter `.` and suggests a command if the user make an input error.
```
genji>.hel
.hel is not a command. You meant: ".help"
genji>.foo 
unknown command ".foo". Enter ".help" for help.
```


cmd/genji/shell/command.go
```diff
@@ -2,20 +2,94 @@ package shell
...
+
+// OrderedBy returns a Sorter that sorts using the less functions, in order.
+// Call its Sort method to sort the data.
+func OrderedBy(less ...lessFunc) *Commands {
+	return &Commands{
+		less: less,
+	}
+}
+
+// Len is part of sort.Interface.
```
asdine, line 66:
```suggestion
// Len implements sort.Interface.
```

cmd/genji/shell/command.go
```diff
@@ -2,20 +2,94 @@ package shell
...
+func (c *Commands) Len() int {
+	return len(c.cmds)
+}
+
+// Swap is part of sort.Interface.
```
asdine, line 71:
```suggestion
// Swap implements sort.Interface.
```

cmd/genji/shell/command.go
```diff
@@ -2,20 +2,94 @@ package shell
...
-	".help":                 "\t\t\tList all commands.",
+type Command struct {
+	Name    string
+	Options string
+	Help    string
+	Action  func(db *genji.DB, cmd []string) error
 }
 
-func runTablesCmd(db *genji.DB, cmd []string) error {
+type lessFunc func(a, b *Command) int
```
asdine, line 21: I don't think we need to go that far for dealing with sorting. There are barely 4 commands we can definitely chose the order manually. Also, I don't think alphabetical order is the best way to sort these commands, I think it would be better to group them by categories in the future.

cmd/genji/shell/command.go
```diff
@@ -92,16 +165,100 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
 
-// runHelpCmd display all available dot commands.
-func runHelpCmd() error {
-	var keys []string
-	for k := range commands {
-		keys = append(keys, k)
+// displayHelp show all available commands.
+func displayHelp(db *genji.DB, cmd []string) error {
+	for _, c := range commands.cmds {
+		indent := 25 - len(c.getOptions()) - len(c.getName())
```
asdine, line 171: What does 25 mean?

cmd/genji/shell/command.go
```diff
@@ -92,16 +165,100 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
 
-	sort.Strings(keys)
-	for _, k := range keys {
-		fmt.Printf("%s %s\n", k, commands[k])
+	return nil
+}
+
+// exitCommand exits this program.
+func exitCommand(db *genji.DB, cmd []string) error {
+	os.Exit(0)
```
asdine, line 181: Let's make sure the db is closed before exiting

cmd/genji/shell/command.go
```diff
@@ -92,16 +165,100 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+
+// getCommandSuggestion returns the suggested command.
+func getCommandSuggestion(in string) Command {
+	suggest := Command{
+		Action: func(db *genji.DB, cmd []string) error {
+			var list []string
+			for _, c := range commands.cmds {
+				name := c.getName()
+				d := levenshtein.ComputeDistance(name, in)
+				if d < 3 {
```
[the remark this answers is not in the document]
tzzed: You are right what do you thing about a distance smaller than the half of the command name. 
`if d < (len(c.Name) / 2)`

cmd/genji/shell/command.go
tzzed: If we don't keep the sort I think is better to delete the `Len()` implementation.

cmd/genji/shell/command.go
tzzed: same as above

cmd/genji/shell/command.go
tzzed: I am agree and I like the "groupby" categories.

cmd/genji/shell/command.go
tzzed: it's the initial number of spaces to keep all descriptions aligned when they are displayed.

cmd/genji/shell/shell.go
```diff
@@ -370,6 +351,21 @@ func (sh *Shell) getAllTables() ([]string, error) {
 }
 
 func (sh *Shell) completer(in prompt.Document) []prompt.Suggest {
+	if strings.HasPrefix(in.Text, ".") {
+		expected := getAllCommands()
+		suggestions := make([]prompt.Suggest, len(expected))
+		for i, e := range expected {
+			suggestions[i].Text = e
+		}
```
asdine, line 359: I think these 5 lines are the same regardless of the query, they should be done once and used here directly instead of being recreated every time a user runs a query.

cmd/genji/shell/shell.go
```diff
@@ -217,32 +218,12 @@ func (sh *Shell) executeInput(in string) error {
 
 func (sh *Shell) runCommand(in string) error {
 	cmd := strings.Fields(in)
-	switch cmd[0] {
```
asdine, line 220: I'm not sure why we don't keep this simple switch instead of doing globals and command types. Do you have a particular reason for pushing this new system?
I think we should focus on keeping things very simple.

cmd/genji/shell/command.go
```diff
@@ -92,16 +101,97 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+	cmdList := make([]string, len(commands))
+	for i := range commands {
+		cmdList[i] = commands[i].Name
+	}
+
+	return cmdList
+}
+
+// getCommandSuggestion returns the suggested command.
+func getCommandSuggestion(in string) Command {
```
asdine, line 147: This doesn't do what the comment suggests, it creates a command that itself determines which command to run. I think this is a bit hacky.

cmd/genji/shell/command.go
asdine: This is marked as resolved but the database is not closed, isn't it?

cmd/genji/shell/command.go
```diff
@@ -2,20 +2,30 @@ package shell
...
-	".help":                 "\t\t\tList all commands.",
+type Command struct {
+	Name    string
+	Options string
+	Help    string
+	Action  func(db *genji.DB, cmd []string) error
 }
 
-func runTablesCmd(db *genji.DB, cmd []string) error {
+const TotalCommands = 4
```
asdine, line 19: Creating globals must be done for a very good reason in my opinion, especially when they are exported.

asdine (changes requested):
I don't think this new `Command` type is worth the trouble, when a simple switch case like we currently have does the job. I believe this PR should simply focus on adding the command suggestions and not redesigning the system, unless there is a very good reason.

cmd/genji/shell/command.go
```diff
@@ -2,20 +2,13 @@ package shell
...
 
-var commands = map[string]string{
-	".tables":               "\t\tList names of tables.",
-	".exit":                 "\t\t\tExit this program.",
-	".indexes [table_name]": "\tDisplay all indexes or the indexes of the given table name.",
-	".help":                 "\t\t\tList all commands.",
-}
-
-func runTablesCmd(db *genji.DB, cmd []string) error {
+// displayTables show all tables.
```
asdine, line 10:
```suggestion
// displayTables shows all tables.
```

cmd/genji/shell/command.go
```diff
@@ -92,17 +85,58 @@ func runIndexesCmd(db *genji.DB, in []string) error {
 	return fmt.Errorf("usage: .indexes [tablename]")
 }
 
-// runHelpCmd display all available dot commands.
-func runHelpCmd() error {
-	var keys []string
-	for k := range commands {
-		keys = append(keys, k)
+// displayHelp show all available cmd.
```
asdine, line 88:
```suggestion
// displayHelp shows all available commands.
```

cmd/genji/shell/command.go
```diff
@@ -92,17 +85,58 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+func displayHelp() error {
+
```
asdine, line 90:
```suggestion
```

cmd/genji/shell/command.go
```diff
@@ -92,17 +85,58 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+		indent := spaces - len(commands[i].options) - len(commands[i].name)
+		fmt.Printf("%s %s%*s%s\n", commands[i].name, commands[i].options, indent, "", commands[i].help)
+	}
+
+	return nil
+}
+
+// displaySuggestion shows suggestions.
+func displaySuggestion(in string) error {
+
```
asdine, line 117:
```suggestion
```

cmd/genji/shell/command.go
```diff
@@ -92,17 +85,58 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+	var suggestions []string
+	commands := []string{".tables", ".indexes", ".help", ".exit"}
+	for _, c := range commands {
+		d := levenshtein.ComputeDistance(c, in)
+		// input should be equal to half command name to get a suggestion.
```
asdine, line 122:
```suggestion
		// input should be equal to half the command name to get a suggestion.
```

cmd/genji/shell/command.go
```diff
@@ -92,17 +85,58 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
 
-	sort.Strings(keys)
-	for _, k := range keys {
-		fmt.Printf("%s %s\n", k, commands[k])
+	fmt.Printf("\"%s\" is not a command. Did you mean: ", in)
+	for i := range suggestions {
+		if i > 0 {
+			fmt.Printf(", ")
+		}
+		fmt.Printf("\"%s\"", suggestions[i])
```
asdine, line 137:
```suggestion
		fmt.Printf("%q", suggestions[i])
```

cmd/genji/shell/command.go
```diff
@@ -2,20 +2,13 @@ package shell
...
 import (
 	"fmt"
-	"sort"
-
+	"github.com/agnivade/levenshtein"
 	"github.com/genjidb/genji"
 	"github.com/genjidb/genji/document"
 )
 
-var commands = map[string]string{
```
asdine, line 11: Before, we had one place where we kept the commands, now we have 3 places we need to update to add them. Why not keeping this map and extract + sort the keys whenever we need them?

cmd/genji/shell/command.go
```diff
@@ -92,17 +85,58 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+	commands = append(commands, command{name: ".indexes",
+		options: "[table name]",
+		help:    "Display all indexes or the indexes of the given table name."})
+
+	commands = append(commands, command{name: ".tables", help: "List names of tables."})
+	commands = append(commands, command{name: ".help", help: "Show available cmd."})
+	commands = append(commands, command{name: ".exit", help: "Exit this program."})
+
+	for i := range commands {
+		// spaces indentation to display helps aligned.
```
asdine, line 106:
```suggestion
		// spaces indentation for readability.
```

cmd/genji/shell/shell.go
```diff
@@ -370,6 +384,16 @@ func (sh *Shell) getAllTables() ([]string, error) {
 }
 
 func (sh *Shell) completer(in prompt.Document) []prompt.Suggest {
+
```
asdine, line 387:
```suggestion
```

Starting a function with a blank line is evil

cmd/genji/shell/command.go
```diff
@@ -92,17 +85,58 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+	for _, c := range commands {
+		d := levenshtein.ComputeDistance(c, in)
+		// input should be equal to half command name to get a suggestion.
+		if d < (len(c) / 2) {
+			suggestions = append(suggestions, c)
+		}
+	}
+
+	if len(suggestions) == 0 {
+		return fmt.Errorf("unknown command %q. Enter \".help\" for help.", in)
```
asdine, line 129:
```suggestion
		return fmt.Errorf("Unknown command %q. Enter \".help\" for help.", in)
```

asdine (changes requested):
I think it could be even simpler, but we're almost there 👍

cmd/genji/shell/command.go
```diff
@@ -2,20 +2,21 @@ package shell
...
 )
 
 var commands = map[string]string{
-	".tables":               "\t\tList names of tables.",
-	".exit":                 "\t\t\tExit this program.",
-	".indexes [table_name]": "\tDisplay all indexes or the indexes of the given table name.",
-	".help":                 "\t\t\tList all commands.",
+	".exit":                 "exit this program",
+	".tables":               "shows all tables",
+	".indexes [table_name]": "shows all indexes",
```
asdine, line 14: Is there a reason why we are changing the command descriptions?

cmd/genji/shell/command.go
```diff
@@ -2,20 +2,21 @@ package shell
...
+	".help":                 "shows help",
 }
 
-func runTablesCmd(db *genji.DB, cmd []string) error {
+// displayTables shows all tables.
+func displayTables(db *genji.DB, cmd []string) error {
```
asdine, line 19: I don't think there is a need to rename this function: we are running the tables command, this is not only displaying something.
Think of it this way: everything the user runs that starts with `.` is a command. However, displaying the suggestion is not a command, that's why we named that function `displaySuggestion`.

cmd/genji/shell/command.go
```diff
@@ -92,17 +93,49 @@ func runIndexesCmd(db *genji.DB, in []string) error {
 	return fmt.Errorf("usage: .indexes [tablename]")
 }
 
-// runHelpCmd display all available dot commands.
-func runHelpCmd() error {
-	var keys []string
+// displayHelp shows all available commands.
+func displayHelp() error {
```
asdine, line 97: Same as above

cmd/genji/shell/command.go
```diff
@@ -92,17 +93,49 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
+
+	return nil
+}
+
+// displaySuggestion shows suggestions.
+func displaySuggestion(in string) error {
+	var suggestions []string
+	for _, c := range commands {
+		d := levenshtein.ComputeDistance(c, in)
+		// input should be equal to the half command name to get a suggestion.
```
asdine, line 120:
```suggestion
		// input should be at least half the command name to get a suggestion.
```

cmd/genji/shell/shell.go
```diff
@@ -123,6 +125,16 @@ func Run(opts *Options) error {
 	return sh.dumpHistory()
 }
 
+func (sh *Shell) registerCommandSuggestion() {
```
asdine, line 128: I don't think we need to register this and keep the state. We should simply rename it `getCommandSuggestions` and return the suggestions based on the `commands` map.
The `expected` slice below is duplicating the commands information which is not ideal, let's just use the global one.

asdine (changes requested):
This is better! Just a few comments, especially regarding renaming decisions. It's definitely simpler like this!

cmd/genji/shell/shell.go
```diff
@@ -370,6 +385,20 @@ func (sh *Shell) getAllTables() ([]string, error) {
 }
 
 func (sh *Shell) completer(in prompt.Document) []prompt.Suggest {
+	if strings.HasPrefix(in.Text, ".") {
+		// This check avoids infinite suggestions.
+		if isCommand(in.Text) {
```
asdine, line 390: This is necessary because you are using `GetWordBeforeCursor` below. If you change it to `in.Text` instead you won't have this problem.

document/array.go
```diff
@@ -8,7 +8,7 @@ import (
 // ErrValueNotFound must be returned by Array implementations, when calling the GetByIndex method and
 // the index wasn't found in the array.
 var (
-	ErrValueNotFound   = errors.New("value not found")
```
asdine, line 11: Please avoid changes that are not related to the PR, this can make rebases more difficult for everyone

document/document_test.go
```diff
@@ -92,9 +92,9 @@ func TestFieldBuffer(t *testing.T) {
 			{"add field", `{"a": {"b": [1, 2, 3]}}`, `c`, document.NewTextValue("foo"), `{"a": {"b": [1, 2, 3]}, "c": "foo"}`, false},
 			{"nested doc", `{"a": "foo"}`, `a`, document.NewDocumentValue(document.NewFieldBuffer().
 				Add("b", document.NewArrayValue(document.NewValueBuffer().
-				Append(document.NewIntegerValue(1)).
-				Append(document.NewIntegerValue(2)).
-				Append(document.NewIntegerValue(3))))), `{"a": {"b": [1, 2, 3]}}`, false},
+					Append(document.NewIntegerValue(1)).
```
asdine, line 95: Same here

asdine (changes requested):
A few comments, but we're almost there 👍🏼

asdine:
Also, please rebase your PR with master, otherwise I can't merge it

cmd/genji/shell/shell.go
```diff
@@ -122,7 +126,24 @@ func Run(opts *Options) error {
 	)
 
 	e.Run()
-	return nil
+
```
asdine, line 129:
```suggestion
```

cmd/genji/shell/shell.go
```diff
@@ -122,7 +126,24 @@ func Run(opts *Options) error {
...
+		err = sh.db.Close()
+		if err != nil {
+			return err
+		}
+	}
+
+	return sh.dumpHistory()
+}
+
+func (sh *Shell) loadCommandSuggestion()  {
```
asdine, line 141:
```suggestion
func (sh *Shell) loadCommandSuggestions()  {
```

cmd/genji/shell/command.go
```diff
@@ -92,17 +97,42 @@ func runIndexesCmd(db *genji.DB, in []string) error {
...
 	}
 
-	sort.Strings(keys)
-	for _, k := range keys {
-		fmt.Printf("%s %s\n", k, commands[k])
+	return nil
+}
+
+// displaySuggestion shows suggestions.
+func displaySuggestion(in string) error {
```
asdine, line 113:
```suggestion
func displaySuggestions(in string) error {
```

asdine (approved):
Some final typos and spacing issues, otherwise everything is good now

asdine (approved):
Thank you @tzzed !

PR #174, opened 2020-08-30
Allow dot notation for UPDATE SET 

### LINKED ISSUE #84: Support dot notation on UPDATE ... SET

tzzed:
This PR fix #84 with the new format of `ValuePath` with change `UpdateSetPair`. It now like this.
```
type updateSetPair struct {
	path document.ValuePath
	e    expr.Expr
}
``` 
More tests was added.

### DIFF
--- a/document/array.go
+++ b/document/array.go
@@ -7,7 +7,10 @@ import (
 
 // ErrValueNotFound must be returned by Array implementations, when calling the GetByIndex method and
 // the index wasn't found in the array.
-var ErrValueNotFound = errors.New("value not found")
+var (
+	ErrValueNotFound   = errors.New("value not found")
+	ErrIndexOutOfRange = errors.New("index out of range")
+)
 
 // An Array contains a set of values.
 type Array interface {
@@ -81,7 +84,7 @@ func (vb ValueBuffer) Iterate(fn func(i int, value Value) error) error {
 // GetByIndex returns a value set at the given index. If the index is out of range it returns an error.
 func (vb ValueBuffer) GetByIndex(i int) (Value, error) {
 	if i >= len(vb) {
-		return Value{}, ErrValueNotFound
+		return Value{}, ErrIndexOutOfRange
 	}
 
 	return vb[i], nil
--- a/document/array.go
+++ b/document/array.go
@@ -9,7 +9,6 @@ import (
 // the index wasn't found in the array.
 var (
 	ErrValueNotFound   = errors.New("value not found")
-	ErrIndexOutOfRange = errors.New("index out of range")
 )
 
 // An Array contains a set of values.
@@ -84,7 +83,7 @@ func (vb ValueBuffer) Iterate(fn func(i int, value Value) error) error {
 // GetByIndex returns a value set at the given index. If the index is out of range it returns an error.
 func (vb ValueBuffer) GetByIndex(i int) (Value, error) {
 	if i >= len(vb) {
-		return Value{}, ErrIndexOutOfRange
+		return Value{}, ErrFieldNotFound
 	}
 
 	return vb[i], nil
--- a/document/document.go
+++ b/document/document.go
@@ -86,16 +86,94 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
 	return Value{}, ErrFieldNotFound
 }
 
+// setFieldValue replaces a field if it already exists or creates one if not.
+func (fb *FieldBuffer) setFieldValue(field string, reqValue Value) error {
+	_, err := fb.GetByField(field)
+	switch err {
+	case ErrFieldNotFound:
+		fb.Add(field, reqValue)
+		return nil
+	case nil:
+		_ = fb.Replace(field, reqValue)
+		return nil
+	}
+
+	return err
+}
+
+// setValueAtPath deep replaces or creates a field
+// using the value path
+func (fb *FieldBuffer) setValueAtPath(v Value, p ValuePath, newValue Value) (Value, error) {
+	switch v.Type {
+	case DocumentValue:
+		var buf FieldBuffer
+		err := buf.ScanDocument(v.V.(Document))
+		if err != nil {
+			return v, err
+		}
+
+		if len(p) == 1 {
+			err = buf.setFieldValue(p[0].FieldName, newValue)
+			return NewDocumentValue(&buf), err
+		}
+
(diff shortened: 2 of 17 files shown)


sql/parser/update.go
```diff
@@ -58,10 +59,12 @@ func (p *Parser) parseSetClause() ([]updateSetPair, error) {
...
 
-		// Scan the identifier for the field name.
-		tok, pos, lit := p.ScanIgnoreWhitespace()
-		if tok != scanner.IDENT {
-			return nil, newParseError(scanner.Tokstr(tok, lit), []string{"identifier"}, pos)
+		// Scan the identifier for the path name.
+		path, err := p.parsePath()
+		if err != nil {
+			pErr := err.(*ParseError)
+			pErr.Expected = []string{"value path"}
```
asdine, line 66:
```suggestion
			pErr.Expected = []string{"path"}
```

sql/planner/optimizer.go
```diff
@@ -252,7 +252,7 @@ func RemoveUnnecessarySelectionNodesRule(t *Tree) (*Tree, error) {
 // UseIndexBasedOnSelectionNodeRule scans the tree for the first selection node whose condition is an
 // operator that satisfies the following criterias:
 // - implements the indexIteratorOperator interface
-// - one of its operands is a field selector that is indexed
+// - one of its operands is a path selector that is indexed
```
asdine, line 255: We are talking about the `expr.FieldSelector` type, perhaps this would be better:

```suggestion
// - one of its operands is a expr.FieldSelector that is indexed
```

sql/planner/projection.go
```diff
@@ -149,7 +149,7 @@ func (r documentMask) MarshalJSON() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-// A ResultField is a field that will be part of the result document that will be returned at the end of a Select statement.
+// A ResultField is a path that will be part of the result document that will be returned at the end of a Select statement.
```
asdine, line 152: `field` is correct here

sql/query/update_test.go
```diff
@@ -77,22 +77,50 @@ func TestUpdateStmt(t *testing.T) {
...
-		require.JSONEq(t, `[{"a": [1, 0, 0], "b": 0}, {"a": [2, 0], "b": 0}]`, buf.String())
+		tests := []struct {
+			name     string
+			query    string
+			fails    bool
+			expected string
+			params   []interface{}
+		}{
+			{"SET / No cond add field ", `UPDATE foo set b = 0`, false, `[{"a": [1, 0, 0], "b": 0}, {"a": [2, 0], "b": 0}]`, nil},
+			{"SET / No cond / with dot notation at existing index only", `UPDATE foo SET a[2] = 10`, false, `[{"a": [1, 0, 10]}, {"a": [2, 0]}]`, nil},
```
asdine, line 88: We should use `path` instead of `dot notation` in these. No need to rename other tests that still have that naming.

document/document.go
```diff
@@ -86,16 +86,94 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+	case nil:
+		_ = fb.Replace(field, reqValue)
+		return nil
+	}
+
+	return err
+}
+
+// setValueAtPath deep replaces or creates a field
+// using the value path
```
asdine, line 105:
```suggestion
// at the given path.
```

document/document.go
```diff
@@ -86,16 +86,94 @@ func (fb FieldBuffer) GetByField(field string) (Value, error) {
...
+func (fb *FieldBuffer) setValueAtPath(v Value, p ValuePath, newValue Value) (Value, error) {
```
asdine, line 106: I think there is no need to make `setValueAtPath` a method of FieldBuffer since it doesn't access it.

asdine (changes requested):
The solution is very elegant, thank you @tzzed!
I only have a few comments, then it will be ready to be merged.
Also please avoid naming your commits the same name

document/array.go
```diff
@@ -81,7 +84,7 @@ func (vb ValueBuffer) Iterate(fn func(i int, value Value) error) error {
 // GetByIndex returns a value set at the given index. If the index is out of range it returns an error.
 func (vb ValueBuffer) GetByIndex(i int) (Value, error) {
 	if i >= len(vb) {
-		return Value{}, ErrValueNotFound
+		return Value{}, ErrIndexOutOfRange
```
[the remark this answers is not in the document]
tzzed: I did it because if statement is if the given index is greater than the `len(vb)`. It seems this the only way to get a nil Value.
```
if i >= len(vb) {
		return Value{}, ErrIndexOutOfRange
	}
```
 And I thought it more pertinent to respond to the following case. 
```
{"index out of range", `{"a": {"b": [1, 2, 3]}}`, `a.b[1000]`, document.NewIntegerValue(1), ``, true},
```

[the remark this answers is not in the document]
asdine: Sorry I don't understand your comment: Is this error used somewhere specific?

sql/parser/parser.go
```diff
@@ -50,6 +50,7 @@ func (p *Parser) ParseQuery() (query.Query, error) {
 			p.Unscan()
 			s, err := p.ParseStatement()
 			if err != nil {
+				fmt.Println("sufgg")
```
asdine, line 53: There is a missing debug line here

sql/planner/optimizer.go
```diff
@@ -252,7 +252,7 @@ func RemoveUnnecessarySelectionNodesRule(t *Tree) (*Tree, error) {
 // UseIndexBasedOnSelectionNodeRule scans the tree for the first selection node whose condition is an
 // operator that satisfies the following criterias:
 // - implements the indexIteratorOperator interface
-// - one of its operands is a expr.FieldSelector that is indexed
+// - one of its operands is the expr.FieldSelectors that is indexed
```
asdine, line 255: Please keep this line as before, as the new one is not correct

asdine (commented):
@tzzed We are almost there! Just 2 little comments and the PR is mergeable! Thanks!

asdine (approved):
Thank you @tzzed 👏

PR #182, opened 2020-09-07
Remove intro text when STDIN comes from a pipe

### LINKED ISSUE #179: Remove introduction text on when reading from STDIN
Currently, whenever we are passing SQL queries to stdin we are also displaying the introduction header:

```bash
$ echo "SELECT 1 + 1" | genji
Opened an in-memory database.
Enter ".help" for usage hints.
{
  "1 + 1": 2
}
```

We want that text to be removed:

```
$ echo "SELECT 1 + 1" | genji
{
  "1 + 1": 2
}
```

ilmanzo:
fix #179

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ profile.*
 
 # Databases
 *.db
+cmd/genji/genji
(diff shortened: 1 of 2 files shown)


asdine (approved):
Thank you @ilmanzo !

PR #187, opened 2020-09-08
Add command insert to cli command

### LINKED ISSUE #185: Add insert command to CLI
The `insert` command allows inserting documents from the command line:

```bash
# inserting json documents to the table foo, as individual arguments
genji insert -f my.db -t foo '{"a": 1}' '{"a": 2}'

# inserting json documents to the table foo, as a json array
genji insert -f my.db -t foo '[{"a": 1}, {"a": 2}]'

# inserting a json stream or json array from stdin
 curl https://api.github.com/repos/genjidb/genji/issues | genji insert -f my.db -t foo
```

The table must have been created beforehand.

Options of the insert command:
```
-f, --file: path of the database file
-t, --table: name of the table, it must already exist
```

tzzed:
Fix #185.

### DIFF
--- a/cmd/genji/insert.go
+++ b/cmd/genji/insert.go
@@ -1,0 +1,145 @@
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"github.com/dgraph-io/badger/v2"
+	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/document"
+	"github.com/genjidb/genji/engine"
+	"github.com/genjidb/genji/engine/badgerengine"
+	"github.com/genjidb/genji/engine/boltengine"
+	"io"
+	"os"
+	"strings"
+)
+
+func skipSpaces(r *bufio.Reader) (byte, error) {
+	var c byte
+	var err error
+
+	for {
+		c, err = r.ReadByte()
+		if err != nil {
+			return c, err
+		}
+
+		if c != '\n' && c != '\r' && c != ' ' && c != '\t' {
+			break
+		}
+	}
+
+	return c, nil
+}
+
+func executeInsertCommand(db *genji.DB, table string, r io.Reader) error {
+	q := fmt.Sprintf("INSERT INTO %s VALUES ?", table)
+	rd := bufio.NewReader(r)
+	var c byte
+	var err error
+
+	// Ignore spaces.
+	c, err = skipSpaces(rd)
+	if err != nil {
+		return err
+	}
+	switch c {
+	case '{': // Json stream
+		if err := rd.UnreadByte(); err != nil {
+			return err
+		}
+
+		dec := json.NewDecoder(rd)
+		for {
+			var fb document.FieldBuffer
+			err := dec.Decode(&fb)
+			if err == io.EOF {
+				break
+			}
+			if err != nil {
+				return err
+			}
+
+			if err := db.Exec(q, &fb); err != nil {
+				return err
+			}
+		}
+
+	case '[': // Array of json objects
+		if err := rd.UnreadByte(); err != nil {
+			return err
+		}
+
+		dec := json.NewDecoder(rd)
+		t, err := dec.Token()
+		if err != nil {
+			return err
+		}
+
+		for dec.More() {
+			var fb document.FieldBuffer
+			err := dec.Decode(&fb)
+			if err != nil && err != io.EOF {
+				return err
+			}
+
+			if err := db.Exec(q, &fb); err != nil {
+				return err
+			}
+
+		}
+
+		t, err = dec.Token()
+		if err != nil {
+			return err
+		}
+		d, ok := t.(json.Delim)
+		if ok && d.String() != "]" {
+			return fmt.Errorf("found %q, but expected ']'", c)
+		}
+
+	default:
+		return fmt.Errorf("found %q, but expected '{' or '['", c)
+
+	}
+
+	return nil
+}
+
+func runInsertCommand(e, DBPath, table string, args []string) error {
+	var ng engine.Engine
+	var err error
+
+	switch e {
+	case "bolt":
+		ng, err = boltengine.NewEngine(DBPath, 0660, nil)
+	case "badger":
+		ng, err = badgerengine.NewEngine(badger.DefaultOptions(DBPath).WithLogger(nil))
+	}
+	if err != nil {
+		return err
+	}
+
+	db, err := genji.New(ng)
+	if err != nil {
+		return err
+	}
+
+	defer db.Close()
+
+	fi, _ := os.Stdin.Stat()
+	m := fi.Mode()
+	// Insert command is given in the pipe
+	if (m & os.ModeNamedPipe) != 0 {
+		return executeInsertCommand(db, table, os.Stdin)
+	}
+
+	for _, arg := range args {
+		if err := executeInsertCommand(db, table, strings.NewReader(arg)); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/cmd/genji/insert.go
+++ b/cmd/genji/insert.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/dgraph-io/badger/v2"
 	"github.com/genjidb/genji"
@@ -14,6 +15,8 @@ import (
 	"os"
 	"strings"
 )
+
+var ErrNoData = errors.New("no data to insert")
 
 func skipSpaces(r *bufio.Reader) (byte, error) {
 	var c byte
@@ -101,7 +104,6 @@ func executeInsertCommand(db *genji.DB, table string, r io.Reader) error {
 
 	default:
 		return fmt.Errorf("found %q, but expected '{' or '['", c)
-
 	}
 
 	return nil
@@ -133,6 +135,10 @@ func runInsertCommand(e, DBPath, table string, args []string) error {
 	// Insert command is given in the pipe
 	if (m & os.ModeNamedPipe) != 0 {
 		return executeInsertCommand(db, table, os.Stdin)
+	}
+
+	if len(args) == 0 {
+		return ErrNoData
 	}
 
 	for _, arg := range args {
--- a/cmd/genji/insert_test.go
+++ b/cmd/genji/insert_test.go
@@ -1,0 +1,136 @@
+package main
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/document"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExecuteInsertCommand(t *testing.T) {
+	tests := []struct {
+		name  string
+		data  string
+		want  string
+		fails bool
+	}{
+		{"Simple Json", `{"a": 1}`, `{"a": 1}`, false},
+		{"JSON object", `{"a": {"b": [1, 2, 3]}}`, `{"a": {"b": [1, 2, 3]}}`, false},
+		{"nested document", `{"a": {"b": [1, 2, 3]}}`, `{"a": {"b": [1, 2, 3]}}`, false},
+		{"nested array multiple indexes", `{"a": {"b": [1, 2, [1, 2, {"c": "foo"}]]}}`, `{"a": {"b": [1, 2, [1, 2, {"c": "foo"}]]}}`, false},
+		{"document in array", `{"a": [{"b":"foo"}, 2, 3]}`, `{"a": [{"b":"foo"}, 2, 3]}`, false},
+		{"Non closed json array", `[{"foo":"bar"}`, ``, true},
+		{"Non closed json stream", `{"foo":"bar"`, ``, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			db, err := genji.Open(":memory:")
+			require.NoError(t, err)
+			defer db.Close()
+
+			err = db.Exec(`CREATE TABLE foo`)
+			require.NoError(t, err)
+			err = executeInsertCommand(db, "foo", strings.NewReader(tt.data))
+			if tt.fails {
+				require.Error(t, err)
+				return
+			}
+
+			require.NoError(t, err)
+			res, err := db.Query("SELECT * FROM foo")
+			defer res.Close()
+			require.NoError(t, err)
+
+			var buf bytes.Buffer
+			err = document.IteratorToJSON(&buf, res)
+			require.NoError(t, err)
+			require.JSONEq(t, tt.want, buf.String())
+
+		})
+	}
+
+	t.Run(`Json Array`, func(t *testing.T) {
+		const jsonArray = `
+	[
+		{"Name": "Ed", "Text": "Knock knock."},
+		{"Name": "Sam", "Text": "Who's there?"},
+		{"Name": "Ed", "Text": "Go fmt."},
+		{"Name": "Sam", "Text": "Go fmt who?"},
+		{"Name": "Ed", "Text": "Go fmt yourself!"}
+	]
+`
+		jsonStreamResult := []string{`{"Name": "Ed", "Text": "Knock knock."}`,
+			`{"Name": "Sam", "Text": "Who's there?"}`, `{"Name": "Ed", "Text": "Go fmt."}`,
+			`{"Name": "Sam", "Text": "Go fmt who?"}`,
+			`{"Name": "Ed", "Text": "Go fmt yourself!"}`}
+
+		db, err := genji.Open(":memory:")
+		require.NoError(t, err)
+		defer db.Close()
+
+		err = db.Exec(`CREATE TABLE foo`)
+		require.NoError(t, err)
+		err = executeInsertCommand(db, "foo", strings.NewReader(jsonArray))
+		require.NoError(t, err)
+		res, err := db.Query("SELECT * FROM foo")
+		defer res.Close()
+		require.NoError(t, err)
+
+		i := 0
+		_ = res.Iterate(func(d document.Document) error {
+			var buf bytes.Buffer
+			err = document.ToJSON(&buf, d)
+			require.NoError(t, err)
+			require.JSONEq(t, jsonStreamResult[i], buf.String())
+			i++
+			return nil
+		})
+	})
+
+	t.Run(`Json Stream`, func(t *testing.T) {
+		const jsonStream = `
+		{"Name": "Ed", "Text": "Knock knock."}
+		{"Name": "Sam", "Text": "Who's there?"}
+		{"Name": "Ed", "Text": "Go fmt."}
+		{"Name": "Sam", "Text": "Go fmt who?"}
+		{"Name": "Ed", "Text": "Go fmt yourself!"}
+		`
+		jsonStreamResult := []string{`{"Name": "Ed", "Text": "Knock knock."}`,
+			`{"Name": "Sam", "Text": "Who's there?"}`, `{"Name": "Ed", "Text": "Go fmt."}`,
+			`{"Name": "Sam", "Text": "Go fmt who?"}`,
+			`{"Name": "Ed", "Text": "Go fmt yourself!"}`}
+
+		db, err := genji.Open(":memory:")
+		defer db.Close()
+		require.NoError(t, err)
+
+		err = db.Exec(`CREATE TABLE foo`)
+		require.NoError(t, err)
+
+		err = executeInsertCommand(db, "foo", strings.NewReader(jsonStream))
+		require.NoError(t, err)
+
+		res, err := db.Query("SELECT * FROM foo")
+		defer res.Close()
+		require.NoError(t, err)
+
+		i := 0
+		_ = res.Iterate(func(d document.Document) error {
+			var buf bytes.Buffer
+			err = document.ToJSON(&buf, d)
+			require.NoError(t, err)
+			require.JSONEq(t, jsonStreamResult[i], buf.String())
+			i++
+			return nil
+		})
+
+		wantCount, err := res.Count()
+		require.NoError(t, err)
+		require.Equal(t, wantCount, i)
+	})
+
+}
--- a/cmd/genji/main.go
+++ b/cmd/genji/main.go
@@ -56,12 +56,22 @@ func main() {
 				engine := c.String("engine")
 				args := c.Args().Slice()
 
-				return runInsertCommand(engine, dbPath, table, args)
+				err := runInsertCommand(engine, dbPath, table, args)
+				switch err {
+				case ErrNoData:
+					cli.ShowAppHelpAndExit(c, 2)
+				case nil:
+					break
+				default:
+					return err
+				}
+
+				return nil
 			},
 
 			OnUsageError: func(c *cli.Context, err error, isSubcommand bool) error {
-				_, _ = fmt.Fprintf(c.App.Writer, "error: %s\n", err.Error())
-				return err
+				cli.ShowAppHelpAndExit(c, 2)
+				return nil
 			},
 		},
 	}
@@ -99,5 +109,6 @@ func main() {
 	err := app.Run(os.Args)
 	if err != nil {
 		_, _ = fmt.Fprintf(os.Stdout, "error: %v\n", err)
+		os.Exit(2)
 	}
 }
--- a/cmd/genji/main.go
+++ b/cmd/genji/main.go
@@ -1,11 +1,11 @@
 package main
 
 import (
-	"log"
+	"fmt"
 	"os"
 
 	"github.com/genjidb/genji/cmd/genji/shell"
-	"github.com/urfave/cli"
+	"github.com/urfave/cli/v2"
 )
 
 func main() {
@@ -24,6 +24,49 @@ func main() {
 		},
 	}
 
+	app.Commands = []*cli.Command{
+		{
+			Name:      "insert",
+			Usage:     "Insert documents from the command line",
+			UsageText: "genji insert [options] [arguments...]",
+			HideHelp:  true,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:    "engine",
+					Aliases: []string{"e"},
+					Usage:   "name of the engine to use, options are 'bolt' or 'badger'. defaults to bolt",
+					Value:   "bolt",
+				},
+				&cli.StringFlag{
+					Name:     "db",
+					Usage:    "path of the database file",
+					Required: true,
+				},
+				&cli.StringFlag{
+					Name:     "table",
+					Aliases:  []string{"t"},
+					Usage:    "name of the table, it must already exist",
+					Required: true,
+				},
+			},
+			Action: func(c *cli.Context) error {
+				dbPath := c.String("db")
+				table := c.String("table")
+				// Use bolt as default engine.
+				engine := c.String("engine")
+				args := c.Args().Slice()
+
+				return runInsertCommand(engine, dbPath, table, args)
+			},
+
+			OnUsageError: func(c *cli.Context, err error, isSubcommand bool) error {
+				_, _ = fmt.Fprintf(c.App.Writer, "error: %s\n", err.Error())
+				return err
+			},
+		},
+	}
+
+	// Root command
 	app.Action = func(c *cli.Context) error {
 		useBolt := c.Bool("bolt")
 		useBadger := c.Bool("badger")
@@ -55,6 +98,6 @@ func main() {
 
 	err := app.Run(os.Args)
 	if err != nil {
-		log.Fatal(err)
+		_, _ = fmt.Fprintf(os.Stdout, "error: %v\n", err)
 	}
 }
--- a/cmd/genji/go.mod
+++ b/cmd/genji/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/mattn/go-tty v0.0.3 // indirect
 	github.com/pkg/term v0.0.0-20200520122047-c3ffed290a03 // indirect
 	github.com/stretchr/testify v1.6.1
-	github.com/urfave/cli v1.22.4
+	github.com/urfave/cli/v2 v2.2.0
 )
 
 replace github.com/genjidb/genji v0.8.0 => ../../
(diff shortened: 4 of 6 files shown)


cmd/genji/main.go
```diff
@@ -33,26 +33,35 @@ func main() {
...
+		var opts *shell.Options
+		var err error
 
-		if (useBolt || useBadger) && dbpath == "" {
-			return cli.NewExitError("db path required when using bolt or badger", 2)
+		// check if there is a dbpath or a command.
+		if len(c.Args()) >= 1 {
+			opts, err = shell.GetOpts(c.Args())
+			if err != nil {
+				fmt.Println(err)
```
asdine, line 43: Except for some very specific reasons, we should never use os.Exit in functions, we should return a `cli.NewExitError` here

asdine (changes requested):
Thanks @tzzed !
You don't need to parse arguments yourself, we are using the https://github.com/urfave/cli CLI framework which takes care of this for us. Please read their documentation to learn how to add a command.
I'll do a proper review after this change 👍

cmd/genji/main.go
tzzed: Using the CLI framework to parse has simplified this part.
We really need to use `log.Fatal(err)` in the `main` which display a datetime:
```
if err := app.Run(os.Args); err != nil {
		log.Fatal(err)
	}

# The output 
2020/09/09 00:05:48 table not found
```

`fmt.Println(err)` doesn't display a datetime.

cmd/genji/main.go
```diff
@@ -26,38 +25,71 @@ func main() {
...
+			},
+			Action: func(c *cli.Context) error {
+				table := fmt.Sprintf("%v", c.Value("table"))
+				dbPath := fmt.Sprintf("%v", c.Value("file"))
+				data := c.Args().Slice()
+				// use bolt as default engine
+				engine := "bolt"
+				opts := &shell.Options{Engine: engine, Data: data, Table: table, DBPath: dbPath}
+				err := shell.RunInsertQuery(opts)
+				return err
```
asdine, line 57:
```suggestion
				return shell.RunInsertQuery(opts)
```

cmd/genji/main.go
```diff
@@ -26,38 +25,71 @@ func main() {
...
+			},
+		},
+	}
+
+	// Use for genji without commands
```
asdine, line 62:
```suggestion
	// Root command
```

cmd/genji/main.go
```diff
@@ -26,38 +25,71 @@ func main() {
...
 
+	app.Commands = []*cli.Command{
+		{
+			Name:      "insert",
+			Usage:     "Allows inserting documents from the command line",
+			UsageText: "genji insert [--file, -f] [db path] [--table, -t] [table name] [arguments...]",
+			HideHelp:  true,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:      "file",
```
asdine, line 36: I realize I made a mistake in the spec, sorry about that.
The way we deal with the database should be the following:

- `--db`: Path of the database
- `--engine, -e`: Name of the engine

cmd/genji/main.go
```diff
@@ -26,38 +25,71 @@ func main() {
...
+					Aliases:  []string{"t"},
+					Usage:    "name of the table, it must already exist",
+					Required: true,
+				},
+			},
+			Action: func(c *cli.Context) error {
+				table := fmt.Sprintf("%v", c.Value("table"))
+				dbPath := fmt.Sprintf("%v", c.Value("file"))
+				data := c.Args().Slice()
+				// use bolt as default engine
```
asdine, line 53: With the comment above, we should now select the right engine

cmd/genji/shell/shell.go
```diff
@@ -44,6 +45,11 @@ type Options struct {
 	Engine string
 	// Path of the database file or directory that will be created.
 	DBPath string
+	// Table insertion table for insert command.
```
asdine, line 48: The shell type is owned by the root command, this is the only command that runs the genji shell.
The insert command shouldn't use the shell at all, it must simply open the selected db, run a bunch of inserts and quit.

document/document_test.go
```diff
@@ -92,13 +91,9 @@ func TestFieldBuffer(t *testing.T) {
 			{"add field", `{"a": {"b": [1, 2, 3]}}`, `c`, document.NewTextValue("foo"), `{"a": {"b": [1, 2, 3]}, "c": "foo"}`, false},
 			{"nested doc", `{"a": "foo"}`, `a`, document.NewDocumentValue(document.NewFieldBuffer().
 				Add("b", document.NewArrayValue(document.NewValueBuffer().
-				Append(document.NewIntegerValue(1)).
-				Append(document.NewIntegerValue(2)).
-				Append(document.NewIntegerValue(3))))), `{"a": {"b": [1, 2, 3]}}`, false},
-			{"nested doc", `{"a": {"b": [1, 2, 3]}}`, `a.b`, document.NewArrayValue(document.NewValueBuffer().
```
asdine, line 98: Could you explain this change? Ideally, we want to avoid changes that are irrelevant to the current PR. If this is important, let's open another PR, even if the change is small it's fine

cmd/genji/shell/shell.go
```diff
@@ -300,6 +355,23 @@ func (sh *Shell) getDB() (*genji.DB, error) {
 	return sh.db, nil
 }
 
+// buildJsonString converts string json array to string json document
+// [{"a":1}, {"b":2}] => {"":[{"a":1}, {"b":2}]}.
+func buildJsonString(data string) string {
```
asdine, line 360: We should not use regexp to transform json.
Use [json.RawMessage](https://golang.org/pkg/encoding/json/#RawMessage) instead:

```go
var array []json.RawMessage
// then decode to that variable
```

cmd/genji/shell/shell.go
```diff
@@ -72,6 +78,55 @@ func stdinFromTerminal() bool {
...
+	if err != nil {
+		return err
+	}
+
+	if ran {
+		return nil
+	}
+
+	if len(opts.Data) == 0 {
+		return fmt.Errorf("error: insert -f %q -t %q %q", sh.opts.DBPath, sh.opts.Table, sh.opts.Data)
```
asdine, line 124: If you want to display the usage, don't do it by hand, use the framework instead

asdine (changes requested):
Sorry I made a mistake in the spec, since we should also support Badger we need to allow passing a directory, see the comments for more info.
Also, we should not use the Shell type for this command, only the root command does this.
The way json is handled could also be improved.
Keep going 💪

cmd/genji/shell/shell.go
tzzed: I see but how to manage the json from stdin?
``` 
$>curl https://api.github.com/repos/genjidb/genji/issues | genji insert -f my.db -t foo
```

document/document_test.go
tzzed: Mistakenly done. Sorry.

cmd/genji/shell/shell.go
tzzed: Going to see.

cmd/genji/main.go
tzzed: What I understand is  what you mean is to replace `--file, -f` by `--db` and keep it required and with no alias.
`--engine, -e` is not required. if `-e` is not specified `bolt` is used a default engine.

cmd/genji/shell/shell.go
tzzed: Even if I use `[]json.RawMessage` I get an error when I insert the datas that contain `\r`. 
`error: found \r, expected } at line 65, char 85`

The only way, I know is to use regexp.
So the `json.RawMessage` keep the best way decode array json.
thanks.

cmd/genji/main.go
asdine: I think it will be better to remove the `--bolt` and `badger` options and replace them by `-e` instead. But not necessarily in this PR

cmd/genji/shell/shell.go
asdine: Reading from stdin is not limited to the Shell type, you can read from stdin yourself. Here is an example: https://sourcegraph.com/github.com/genjidb/genji@master/-/blob/cmd/genji/shell/shell.go?utm_source=VSCode-1.1.0#L304-314

cmd/genji/shell/shell.go
asdine: Can you provide an example here?

tzzed: It happens when I run.
`curl https://api.github.com/repos/genjidb/genji/issues | genji insert -f my.db -t foo`

asdine: Can you give me more details please? Like what piece of code is reading the data. Ideally, write a simplified example in a comment here with juste the relevant code and input.

tzzed:
```
# reads input
in, err := ioutil.ReadAll(os.Stdin)
	if err != nil {
		return fmt.Errorf("unable to read piped input: %w", err)
	}

return executeInsertCommand(db, table, string(in))
```

```
# here is executed the query and got error.
func executeInsertCommand(db *genji.DB, table, data string) error {
	var b strings.Builder
	/*
         # filing b with INSERT INTO table VALUES 
       */
var jRaw []json.RawMessage
	if err := json.Unmarshal([]byte(data), &jRaw); err != nil {
		return err
	}

	for i := range jRaw {
		b.WriteString(string(jRaw[i]))
		err := db.Exec(b.String())
		if err != nil {
                     # getting error from db.Exec()
			return err
		}
		// Reset the buffer and fill it with new insert.
		b.Reset()
		b.WriteString(insert)
	}

  return nil
}
```

asdine: You should not try to build a string for a query, use query parameters instead. Maybe json.RawMessage doesn't work here, try using a map:

```go
var m map[string]interface{}
json.Unmarshal(&m)
db.Exec("INSERT INTO foo VALUES ?", document.NewFromMap(&m))
```

cmd/genji/main.go
tzzed: Got it. 
So replace `--bolt` and `--badger` by `-e` in a new PR.

cmd/genji/shell/command.go
```diff
@@ -106,3 +107,40 @@ func runHelpCmd() error {
 
 	return nil
 }
+
+// executeInsertCommand insert json data.
+func executeInsertCommand(db *genji.DB, table string, data []byte) error {
+	query := fmt.Sprintf("INSERT INTO %s VALUES ? ", table)
+	if data[0] != '[' {
```
asdine, line 114: There are basically two ways of receiving json documents:
- JSON stream
- JSON array

### JSON stream

The input will look like this:

```json
{"a": 1}
{"a": 2}
...
```

Basically, you will have a json objects separated by newlines (but they also can contain newlines inside them).
To parse this, you must use [`json.NewDecoder()`](https://golang.org/pkg/encoding/json/#Decoder).
This is also more efficient because the stream of json can be very large and we don't want to load everything in memory before inserting.

## JSON array

The input will look like this:

```json
[
{"a": 1},
{"a": 2}
]
```

This is a proper JSON array, if we use the [`json.NewDecoder()`](https://golang.org/pkg/encoding/json/#Decoder) Decode method directly, this will load the entire array in memory, we don't want to do that.
Instead we're gonna have to decode semi-manually, using the [Token ](https://golang.org/pkg/encoding/json/#Decoder.Token) method. The good thing, the documentation has an example that [exactly that](https://golang.org/pkg/encoding/json/#Decoder.Decode).

cmd/genji/main.go
```diff
@@ -26,38 +25,71 @@ func main() {
 		},
 	}
 
+	app.Commands = []*cli.Command{
+		{
+			Name:      "insert",
+			Usage:     "Allows inserting documents from the command line",
```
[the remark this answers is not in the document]
asdine: This is marked as resolved but wasn't applied

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
+			UsageText: "genji insert [options] [arguments...]",
+			HideHelp:  true,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:     "engine",
+					Aliases:  []string{"e"},
+					Usage:    "Store engine bolt or badger",
```
asdine, line 38:
```suggestion
					Usage:    "name of the engine to use, options are 'bolt' or 'badger'. defaults to bolt",
```

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
+				},
+				&cli.StringFlag{
+					Name:     "table",
+					Aliases:  []string{"t"},
+					Usage:    "name of the table, it must already exist",
+					Required: true,
+				},
+			},
+			Action: func(c *cli.Context) error {
+				dbPath := fmt.Sprintf("%v", c.Value("db"))
```
asdine, line 53: This feels like a hack, the framework provides methods for getting the flags with the right type: https://godoc.org/github.com/urfave/cli#Context.String
```suggestion
				dbPath := c.String("db")
```

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
+				table := fmt.Sprintf("%v", c.Value("table"))
```
asdine, line 54:
```suggestion
				table := c.String("table")
```

cmd/genji/main.go
asdine: You should use the `Value` field to setup the default value: https://godoc.org/github.com/urfave/cli#StringFlag

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
+				// Use bolt as default engine.
+				engine := "bolt"
+				if c.IsSet("engine") {
+					engine = fmt.Sprintf("%v", c.Value("engine"))
```
asdine, line 58:
```suggestion
					engine = c.String("engine")
```

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
+				}
+
+				data := c.Args().Slice()
+				err := shell.RunInsertCommand(engine, dbPath, table, data)
+				// Print error but return nil to avoid double error printing.
```
asdine, line 63: Why not just returning the error and let the framework print it?

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
+					_, _ = fmt.Fprintf(c.App.Writer, "error: %s\n", err.Error())
+					_ = cli.NewExitError(cli.ShowSubcommandHelp(c), 0)
+					return nil
+				}
+
+				return nil
+			},
+
+			OnUsageError: func(c *cli.Context, err error, isSubcommand bool) error {
+				_, _ = fmt.Fprintf(c.App.Writer, "error: %s\n", err.Error())
```
asdine, line 74: Same as above

cmd/genji/main.go
asdine: This is marked as resolved but wasn't applied

cmd/genji/shell/shell.go
asdine: This is marked as resolved by wasn't applied. Please only resolve conversations when they are applied to the PR

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
-			DBPath: dbpath,
-		})
+		return shell.Run(opts)
 	}
 
-	err := app.Run(os.Args)
-	if err != nil {
-		log.Fatal(err)
+	if err := app.Run(os.Args); err != nil {
+		fmt.Println(err)
```
asdine, line 112: Is there a reason we are printing the error this way?

cmd/genji/main.go
```diff
@@ -26,38 +25,91 @@ func main() {
...
+				if c.IsSet("engine") {
+					engine = fmt.Sprintf("%v", c.Value("engine"))
+				}
+
+				data := c.Args().Slice()
+				err := shell.RunInsertCommand(engine, dbPath, table, data)
+				// Print error but return nil to avoid double error printing.
+				if err != nil {
+					_, _ = fmt.Fprintf(c.App.Writer, "error: %s\n", err.Error())
+					_ = cli.NewExitError(cli.ShowSubcommandHelp(c), 0)
```
asdine, line 66: We should not display the usage if the error comes from genji itself

cmd/genji/shell/shell.go
```diff
@@ -72,6 +77,49 @@ func stdinFromTerminal() bool {
...
+			if err != nil {
+				return err
+			}
+		}
+
+		return nil
+	}
+
+	// Insert command is given in the pipe
+	in, err := ioutil.ReadAll(os.Stdin)
```
asdine, line 115: We should not read all stdin because it can get very big, `os.Stdin` is an `io.Reader`, `json.NewDecoder` takes an `io.Reader`, we should simply pass it as is. This means that `executeInsertCommand` should expect a reader, not a byte slice

document/document_test.go
asdine: This is marked as resolved but wasn't applied

cmd/genji/shell/shell.go
```diff
@@ -72,6 +77,49 @@ func stdinFromTerminal() bool {
...
+
+	if err != nil {
+		return err
+	}
+
+	db, err := genji.New(ng)
+	if err != nil {
+		return err
+	}
+
```
asdine, line 100:
```suggestion
```

cmd/genji/shell/shell.go
```diff
@@ -72,6 +77,49 @@ func stdinFromTerminal() bool {
...
+	defer db.Close()
```
asdine, line 101:
```suggestion
	defer db.Close()

```

cmd/genji/shell/command.go
```diff
@@ -106,3 +107,70 @@ func runHelpCmd() error {
...
+		}
+
+		// Check the array is correctly closed.
+		if d, ok := t.(json.Delim); ok && d.String() != "]" {
+			return fmt.Errorf("found %q, expected ']'", d.String())
+		}
+
+		// Json format is valid.
+		// Convert map to a document.
+		d, err := document.NewFromMap(m)
```
asdine, line 150: IIRC, you used [`document.NewFromJSON`](https://sourcegraph.com/github.com/genjidb/genji/-/blob/document/json.go#L16:6) at some point which was a good idea, much better than what I proposed initially, any reason why you are now using `NewFromMap` instead?

cmd/genji/shell/command.go
```diff
@@ -106,3 +107,70 @@ func runHelpCmd() error {
...
+			return err
+		}
+
+		err = db.Exec(q, &d)
+		if err != nil {
+			return err
+		}
+	case '{':
+		var fb document.FieldBuffer
+		err := fb.UnmarshalJSON(data)
```
asdine, line 161: Same here, [`document.NewFromJSON`](https://sourcegraph.com/github.com/genjidb/genji/-/blob/document/json.go#L16:6) does this

cmd/genji/shell/command.go
```diff
@@ -106,3 +107,70 @@ func runHelpCmd() error {
...
+	}
+
+	d, ok := t.(json.Delim)
+	// check if the token is a json delimiter.
+	if !ok {
+		return fmt.Errorf("found %q, expected '{' or '['", d.String())
+	}
+
+	switch d {
+	case '[': // Array json stream
```
asdine, line 127:
```suggestion
	case '[': // Array of json objects
```

cmd/genji/shell/command.go
```diff
@@ -106,3 +107,70 @@ func runHelpCmd() error {
...
+		var m map[string]interface{}
+		for dec.More() {
+			err := dec.Decode(&m)
```
asdine, line 130: All the objects will be merged into the same object, this is not what we want. We want separate objects:

```bash
 genji insert -t wesh --db wesh.db '[{"a": 1},{"a": 2}]'
```

Currently:
```sql
genji> SELECT * FROM wesh;
{
  "a": 2
}
```

Expected:
```sql
genji> SELECT * FROM wesh;
{
  "a": 1
}
{
  "a": 2
}
```

cmd/genji/shell/command_test.go
```diff
@@ -81,3 +81,56 @@ func TestIndexesCmd(t *testing.T) {
...
+func TestExecuteInsertCommand(t *testing.T) {
+	const jsonStream = `
+	{"Name": "Ed", "Text": "Knock knock."}
+	{"Name": "Sam", "Text": "Who's there?"}
+	{"Name": "Ed", "Text": "Go fmt."}
+	{"Name": "Sam", "Text": "Go fmt who?"}
+	{"Name": "Ed", "Text": "Go fmt yourself!"}
+`
+	const jsonArray = `[
+
```
asdine, line 94:
```suggestion
```

cmd/genji/shell/command_test.go
```diff
@@ -81,3 +81,56 @@ func TestIndexesCmd(t *testing.T) {
...
+		{"Name": "Ed", "Text": "Go fmt yourself!"}
+	]`
+	tests := []struct {
+		name   string
+		data   []byte
+		fails  bool
+	}{
+		{"Simple document", []byte(`{"foo":"bar"}`), false},
+		{"multiple documents", []byte(`{"foo":"bar"},{"bar": "foo"}, {"baz":"baz"}`), false},
+		{"Array json", []byte(`[{"name": "Poli","designation": "Police car"},{"name": "Roy","designation": "Fire truck"}]`), false},
```
asdine, line 108:
```suggestion
		{"JSON array", []byte(`[{"name": "Poli","designation": "Police car"},{"name": "Roy","designation": "Fire truck"}]`), false},
```

cmd/genji/shell/command_test.go
```diff
@@ -81,3 +81,56 @@ func TestIndexesCmd(t *testing.T) {
...
+		{"document json", []byte(`{"a": {"b": [1, 2, 3]}}`), false},
```
asdine, line 109:
```suggestion
		{"JSON object", []byte(`{"a": {"b": [1, 2, 3]}}`), false},
```

asdine (changes requested):
Thanks @tzzed ! There are still some changes of the previous review that weren't applied, it is better if you don't mark them as resolved or ask for a review until they are done.

## Shell

We should move the code you added from the `shell` package. Ideally, this PR should not change that package at all as it has nothing to do with this new feature.
The shell package manages the shell feature, meaning the feature that displays a prompt and executes SQL queries. It is not supposed to deal with anything else.

## CLI configuration

The urface/cli framework provides error handling, type-safe flags, default values, etc. we should use these features.

cmd/genji/shell/shell.go
tzzed: I thought about it afterwards.

document/document_test.go
tzzed: Now is done. Sorry

cmd/genji/main.go
tzzed: It was before using cli. I did it to delete the datetime.

cmd/genji/main.go
tzzed: If we return the error, it will be print by `log.Fatal(err)` and the datetime is printed. I find it weird. So what do you think? Should I return `err` and use `log.Fatal(err)` ?

asdine: I understand, in that case how about replac
ing `log.Fatal` by `fmt.Fprintln(os.Stderr, err)`

cmd/genji/shell/shell.go
tzzed: I just check if it is in a pipe with the following
```go

        fi, _ := os.Stdin.Stat()
	m := fi.Mode()
	//Insert command is given in the pipe
	if (m & os.ModeNamedPipe) != 0 {
		return executeInsertCommand(db, table, os.Stdin)
	}
```

cmd/genji/insert.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	switch c {
+	case '{': // Json objects
+		if err := rd.UnreadByte(); err != nil {
+			return err
+		}
+
+		dec := json.NewDecoder(rd)
+		// NewFromDecoder is a wrapper to call parseJsonDocument.
+		d, err := document.NewFromDecoder(dec)
```
asdine, line 56: This is not necessary, document.FieldBuffer has an [UnmarshalJSON ](https://pkg.go.dev/github.com/genjidb/genji@v0.7.1/document#FieldBuffer.UnmarshalJSON)method, you can simply decode to it

```go
var fb document.FieldBuffer

err := dec.Decode(&fb)

db.Exec(&fb)
```

cmd/genji/insert.go
```diff
@@ -0,0 +1,140 @@
...
+	var c byte
+	var err error
+
+	// Ignore spaces.
+	c, err = skipSpaces(rd)
+	if err != nil {
+		return err
+	}
+	switch c {
+	case '{': // Json objects
```
asdine, line 49:
```suggestion
	case '{': // json stream
```

asdine, line 49: This only works for one json object, a json stream is a **multiple** json objects, you need to look until the reader is over. You have an example here: https://golang.org/pkg/encoding/json/#Decoder

cmd/genji/insert.go
```diff
@@ -0,0 +1,140 @@
...
+			return err
+		}
+	case '[': // Array of json objects
+		if err := rd.UnreadByte(); err != nil {
+			return err
+		}
+
+		dec := json.NewDecoder(rd)
+		// Unmarshall json returns []interface{} for JSON arrays.
+		var m []interface{}
```
asdine, line 71: This will load every json document in memory. To give you an example, imagine inserting a json file of 100 GB and passing it to this program like this:

```bash
cat veryBigFile.json | genji insert ...
```

The current code will read all the file content into `m`, and unless we have 100 gb ram computers, the program will crash.
Instead, we want to read chunk by chunk, to keep the memory usage very low, and the json decoder does exactly that.
You have exactly what you need in this example, just copy paste it and adapt it to this function: https://golang.org/pkg/encoding/json/#Decoder

cmd/genji/insert.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	db, err := genji.New(ng)
+	if err != nil {
+		return err
+	}
+
```
asdine, line 122:
```suggestion
```

cmd/genji/insert.go
```diff
@@ -0,0 +1,140 @@
...
+	var ng engine.Engine
+	var err error
+
+	switch e {
+	case "bolt":
+		ng, err = boltengine.NewEngine(DBPath, 0660, nil)
+	case "badger":
+		ng, err = badgerengine.NewEngine(badger.DefaultOptions(DBPath).WithLogger(nil))
+	}
+
```
asdine, line 113: Always group related code together, blank lines should be used to divide logical groups
```suggestion
```

cmd/genji/insert.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	default:
+		return fmt.Errorf("found %q, but expected '{' or '['", c)
+
+	}
+
+	return nil
+}
+
+func runInsertCommand(e, DBPath, table string, data []string) error {
```
asdine, line 103: `data` is too broad here, better name this `args`

cmd/genji/insert_test.go
```diff
@@ -0,0 +1,130 @@
...
+		i := 0
+		_ = res.Iterate(func(d document.Document) error {
+			var buf bytes.Buffer
+			err = document.ToJSON(&buf, d)
+			require.NoError(t, err)
+			require.JSONEq(t, jsonStreamResult[i], buf.String())
+			i++
+			return nil
+		})
+
```
asdine, line 127: This test doesn't check `i`, it is possible that the inner function is never called so `i` must be checked after the call to iterate

document/json.go
```diff
@@ -72,6 +72,19 @@ func (v Value) String() string {
 	return string(d)
 }
 
+// NewFromDecoder creates a document from a json object json.Decoder
+func NewFromDecoder(dec *json.Decoder) (Document, error) {
```
asdine, line 76: This function is not necessary, see comment above.
Basically, [FieldBuffer ](https://pkg.go.dev/github.com/genjidb/genji@v0.7.1/document#FieldBuffer) implements the right interfaces ([json.Marshaler](https://golang.org/pkg/encoding/json/#Marshaler) and [json.Unmarshaler](https://golang.org/pkg/encoding/json/#Unmarshaler)) which makes it compatible with all the json related functions.

asdine (changes requested):
Mostly comments about stream handling and a few errors. Keep up the good work 💪🏼 !

cmd/genji/insert.go
tzzed: I love this solution 👍

cmd/genji/main.go
tzzed: It what I did, I replaced `log.Fatal` by `fmt.Println(err)` but I don't told you why. Should I change it, by `fmt.Fprintln(os.stderr, err)`?

cmd/genji/main.go
asdine: What's the benefit of using this? Wouldn't it display the error twice?

cmd/genji/main.go
```diff
@@ -57,7 +100,7 @@ func main() {
 
 	err := app.Run(os.Args)
 	if err != nil {
-		log.Fatal(err)
+		_, _ = fmt.Fprintf(os.Stdout, "error: %v\n", err)
```
asdine, line 103:
```suggestion
		_, _ = fmt.Fprintf(os.Stdout, "error: %v\n", err)
		os.Exit(2)
```

cmd/genji/insert.go
```diff
@@ -0,0 +1,145 @@
...
+			return err
+		}
+		d, ok := t.(json.Delim)
+		if ok && d.String() != "]" {
+			return fmt.Errorf("found %q, but expected ']'", c)
+		}
+
+	default:
+		return fmt.Errorf("found %q, but expected '{' or '['", c)
+
```
asdine, line 104:
```suggestion
```

cmd/genji/insert.go
```diff
@@ -0,0 +1,145 @@
...
+	defer db.Close()
+
+	fi, _ := os.Stdin.Stat()
+	m := fi.Mode()
+	// Insert command is given in the pipe
+	if (m & os.ModeNamedPipe) != 0 {
+		return executeInsertCommand(db, table, os.Stdin)
+	}
+
+	for _, arg := range args {
```
asdine, line 138: If no arguments are passed, nothing will happen. We need to ensure that if we are not reading from stdin, the user must pass at least one argument

tzzed: Ok, what do you think about `errors.New("no data to insert ")`

asdine: I think displaying the usage should be enough in this case

asdine (approved):
Thank you!

PR #200, opened 2020-09-25
.dump command 

### LINKED ISSUE #181: Add .dump command
The `.dump` command displays the database content as SQL queries. By default, it outputs all the database but it can also display only one table.

Let's say we have a database with multiple tables and indexes, here is a commented example of the expected output

```sql
genji> .dump
-- the dumped script must always be wrapped in a transaction
BEGIN TRANSACTION;
-- each table must be outputted one by one, with table creation first
-- then index creation
-- then inserts
CREATE TABLE foo;
CREATE INDEX idx_foo_a ON foo(a);
-- inserts should use the document notation, as it should be simpler to just encode the document to json and paste it after each VALUES clause
INSERT INTO foo VALUES {"a": 1};
-- DOUBLES must be outputted with a decimal point
INSERT INTO foo VALUES {"a": 2.0, "c": 3.1};
-- ...
-- blank line between tables

CREATE TABLE bar(
  a INTEGER PRIMARY KEY,
  b.c[0] DOUBLE BOT NULL
);
-- ...
COMMIT;
```

tzzed:
This PR fix #181.
All database is displayed or the given table. 
Double are now outputted even with `.0` decimal

### DIFF
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -152,8 +152,9 @@ func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
 		return err
 	}
 
-	c := fmt.Sprintf("CREATE TABLE %s", t.Name())
-	buf.WriteString(c)
+	if _, err = fmt.Fprintf(w, "CREATE TABLE %s", t.Name()); err != nil {
+		return err
+	}
 
 	ti, err := t.Info()
 	if err != nil {
@@ -185,19 +186,18 @@ func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
 
 	// Fields constraints close parenthesis.
 	if len(fcs) > 0 {
-		buf.WriteString("\n);")
+		buf.WriteString("\n);\n")
 	} else {
-		buf.WriteString(";")
+		buf.WriteString(";\n")
 	}
 
 	// Print CREATE TABLE statement.
-	_, err = fmt.Fprintln(w, buf.String())
-	if err != nil {
+	if _, err = buf.WriteTo(w); err != nil {
 		return err
 	}
 	buf.Reset()
 
-	// Indexes statement.
+	// Indexes statements.
 	indexes, err := t.Indexes()
 	if err != nil {
 		return err
@@ -209,9 +209,8 @@ func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
 			u = " UNIQUE"
 		}
 
-		info := fmt.Sprintf("CREATE%s INDEX %s ON %s (%s);", u, index.Opts.IndexName, index.Opts.TableName,
+		_, err = fmt.Fprintf(w, "CREATE%s INDEX %s ON %s (%s);\n", u, index.Opts.IndexName, index.Opts.TableName,
 			index.Opts.Path)
-		_, err = fmt.Fprintln(w, info)
 		if err != nil {
 			return err
 		}
@@ -234,9 +233,9 @@ func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
 			return err
 		}
 		buf.Write(data)
-		buf.WriteString(";")
+		buf.WriteString(";\n")
 
-		if _, err = fmt.Fprintln(w, buf.String()); err != nil {
+		if _, err = buf.WriteTo(w); err != nil {
 			return err
 		}
 
@@ -246,33 +245,33 @@ func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
 	})
 }
 
-// runDumpCmd run .dump command where tables slice is the given tables or it can be empty that is consider all database.
+// runDumpCmd dumps the given tables if provided, otherwise it dumps the whole database.
 func runDumpCmd(db *genji.DB, tables []string, w io.Writer) error {
 	tx, err := db.Begin(false)
 	if err != nil {
 		return err
 	}
 	defer tx.Rollback()
-	_, err = fmt.Fprintln(w, "BEGIN TRANSACTION;")
-	if err != nil {
+
+	if _, err = fmt.Fprintln(w, "BEGIN TRANSACTION;"); err != nil {
 		return err
 	}
 
-	i := 0
-	for _, table := range tables {
+	for i, table := range tables {
 		err = dumpTable(tx, table, w)
 		switch err {
 		case nil:
 			// Blank separation between tables.
 			if i > 0 {
-				fmt.Println()
+				if _, err := fmt.Fprintln(w, ""); err != nil {
+					return err
+				}
 			}
-			i++
 
 		case database.ErrTableNotFound: // If table doesn't exist we skip it.
 			continue
 		default:
-			_, err = fmt.Fprintln(w, "ROLLBACK;")
+			_, err = fmt.Fprintln(w, "COMMIT;")
 			return err
 		}
 	}
@@ -293,11 +292,13 @@ func runDumpCmd(db *genji.DB, tables []string, w io.Writer) error {
 	}
 	defer res.Close()
 
-	i = 0
+	i := 0
 	err = res.Iterate(func(d document.Document) error {
 		// Blank separation between tables.
 		if i > 0 {
-			fmt.Println()
+			if _, err := fmt.Fprintln(w, ""); err != nil {
+				return err
+			}
 		}
 		i++
 
@@ -318,6 +319,6 @@ func runDumpCmd(db *genji.DB, tables []string, w io.Writer) error {
 		return err
 	}
 
-	_, err = fmt.Fprintln(w, "ROLLBACK;")
+	_, err = fmt.Fprintln(w, "COMMIT;")
 	return err
 }
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -1,7 +1,12 @@
 package shell
 
 import (
+	"bytes"
 	"fmt"
+	"os"
+	"strings"
+
+	"github.com/genjidb/genji/database"
 
 	"github.com/agnivade/levenshtein"
 	"github.com/genjidb/genji"
@@ -17,6 +22,7 @@ var commands = []struct {
 	{".help", ``, "List all commands."},
 	{".tables", ``, "List names of tables."},
 	{".indexes", `[table_name]`, "Display all indexes or the indexes of the given table name."},
+	{".dump", `[table_name]`, "Dump database content or table content as SQL statements"},
 }
 
 // runTablesCmd shows all tables.
@@ -134,5 +140,195 @@ func displaySuggestions(in string) error {
 	}
 
 	fmt.Println()
+	return nil
+}
+
+// dumpTable displays the content of the given table as SQL statements.
+func dumpTable(tx *genji.Tx, t *database.Table) error {
+	var buf bytes.Buffer
+
+	c := fmt.Sprintf("CREATE TABLE %s", t.Name())
+	buf.WriteString(c)
+
+	ti, err := t.Info()
+	if err != nil {
+		return err
+	}
+
+	fcs := ti.FieldConstraints
+	hasField := false
+	for i, fc := range fcs {
+		// Fields constraints should be displaying between parenthesis.
+		if !hasField {
+			buf.WriteString("(\n")
+			hasField = true
+		}
+
+		// Don't put the last comma.
+		if i > 0 {
+			buf.WriteString(",\n")
+		}
+
+		buf.WriteString("  " + fcs[i].Path.String() + " ")
+		buf.WriteString(strings.ToUpper(fcs[i].Type.String()))
+		if fc.IsPrimaryKey {
+			buf.WriteString(" PRIMARY KEY")
+		}
+
+		if fc.IsNotNull {
+			buf.WriteString(" NOT NULL")
+		}
+	}
+
+	// Fields constraints close parenthesis.
+	if hasField {
+		buf.WriteString("\n);\n")
+	} else {
+		buf.WriteString(";\n")
+	}
+
+	// Print CREATE TABLE statement.
+	_, err = fmt.Fprintf(os.Stdout, buf.String())
+	if err != nil {
+		return err
+	}
+	buf.Reset()
+
+	// Indexes statement.
+	indexes, err := t.Indexes()
+	if err != nil {
+		return err
+	}
+
+	for _, index := range indexes {
+		u := ""
+		if index.Opts.Unique {
+			u = " UNIQUE"
+		}
+
+		info := fmt.Sprintf("CREATE%s INDEX %s ON %s (%s);\n", u, index.Opts.IndexName, index.Opts.TableName,
+			index.Opts.Path)
+		buf.WriteString(info)
+
+		_, err = fmt.Fprintf(os.Stdout, buf.String())
+		if err != nil {
+			return err
+		}
+		buf.Reset()
+	}
+
+	q := fmt.Sprintf("SELECT * FROM %s", t.Name())
+	res, err := tx.Query(q)
+	if err != nil {
+		return err
+	}
+	defer res.Close()
+
+	// Inserts statement.
+	insert := fmt.Sprintf("INSERT INTO %s VALUES ", t.Name())
+	return res.Iterate(func(d document.Document) error {
+		buf.WriteString(insert)
+
+		data, err := document.MarshalJSON(d)
+		if err != nil {
+			return err
+		}
+		buf.Write(data)
+		buf.WriteString(";\n")
+
+		if _, err = fmt.Fprintf(os.Stdout, buf.String()); err != nil {
+			return err
+		}
+
+		buf.Reset()
+
+		return nil
+	})
+}
+
+// runDumpCmd run .dump command where tables slice is the given tables or it can be empty that is consider all database.
+func runDumpCmd(db *genji.DB, tables []string) error {
+	tx, err := db.Begin(false)
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
+
+	fmt.Println("BEGIN TRANSACTION;")
+
+	// Dump the given table(s) content.
+	argsTable := false
+	i := 0
+	for _, table := range tables {
+		argsTable = true
+		t, err := tx.GetTable(table)
+		switch err {
+		case nil:
+			// Blank separation between tables.
+			if i > 0 {
+				fmt.Println()
+			}
+			i++
+
+			if err := dumpTable(tx, t); err != nil {
+				fmt.Println("ROLLBACK;")
+				return err
+			}
+		case database.ErrTableNotFound: // If table doesn't exist we skip it.
+			break
+		default:
+			fmt.Println("ROLLBACK;")
+			return err
+		}
+	}
+
+	// tables slice argument is not empty, all args tables has been displayed.
+	// If it is empty we should print the database content.
+	if argsTable {
+		fmt.Println("COMMIT;")
+		return nil
+	}
+
+	// tables slice argument is empty.
+	// Dump database content.
+	res, err := tx.Query("SELECT table_name FROM __genji_tables")
+	if err != nil {
+		fmt.Println("COMMIT;")
+		return err
+	}
+	defer res.Close()
+
+	i = 0
+	err = res.Iterate(func(d document.Document) error {
+		// Blank separation between tables.
+		if i > 0 {
+			fmt.Println()
+		}
+		i++
+
+		// Get table name.
+		var tableName string
+		if err := document.Scan(d, &tableName); err != nil {
+			return err
+		}
+
+		t, err := tx.GetTable(tableName)
+		if err != nil {
+			return err
+		}
+
+		if err := dumpTable(tx, t); err != nil {
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
+		fmt.Println("ROLLBACK;")
+		return err
+	}
+
+	fmt.Println("COMMIT;")
+
 	return nil
 }
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -3,7 +3,7 @@ package shell
 import (
 	"bytes"
 	"fmt"
-	"os"
+	"io"
 	"strings"
 
 	"github.com/genjidb/genji/database"
@@ -144,8 +144,13 @@ func displaySuggestions(in string) error {
 }
 
 // dumpTable displays the content of the given table as SQL statements.
-func dumpTable(tx *genji.Tx, t *database.Table) error {
+func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
 	var buf bytes.Buffer
+
+	t, err := tx.GetTable(tableName)
+	if err != nil {
+		return err
+	}
 
 	c := fmt.Sprintf("CREATE TABLE %s", t.Name())
 	buf.WriteString(c)
@@ -156,15 +161,13 @@ func dumpTable(tx *genji.Tx, t *database.Table) error {
 	}
 
 	fcs := ti.FieldConstraints
-	hasField := false
-	for i, fc := range fcs {
-		// Fields constraints should be displaying between parenthesis.
-		if !hasField {
-			buf.WriteString("(\n")
-			hasField = true
-		}
+	// Fields constraints should be displayed between parenthesis.
+	if len(fcs) > 0 {
+		buf.WriteString(" (\n")
+	}
 
-		// Don't put the last comma.
+	for i, fc := range fcs {
+		// Don't display the last comma.
 		if i > 0 {
 			buf.WriteString(",\n")
 		}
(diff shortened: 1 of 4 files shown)


tzzed:
Due to the command `.dump` we need to know if the index creation need `IF NOT EXIST`? 
I think we should add it into `database.IndexConfig struct`...

asdine:
@tzzed We don't need to know that information actually, we don't need to output `IF NOT EXISTS`

cmd/genji/shell/command.go
```diff
@@ -17,6 +21,7 @@ var commands = []struct {
 	{".help", ``, "List all commands."},
 	{".tables", ``, "List names of tables."},
 	{".indexes", `[table_name]`, "Display all indexes or the indexes of the given table name."},
+	{".dump", `[table_name]`, "Display all the database or the given table."},
```
asdine, line 24:
```suggestion
	{".dump", `[table_name]`, "Dump database content or table content as SQL statements"},
```

document/json.go
```diff
@@ -35,6 +35,17 @@ func (v Value) MarshalJSON() ([]byte, error) {
 		x = string(v.V.(string))
 	case DurationValue:
 		x = v.V.(time.Duration).String()
+	case DoubleValue:
```
asdine, line 38: This is a good catch, but this was already fixed in another PR, you should ignore this for the moment

cmd/genji/shell/command.go
```diff
@@ -136,3 +141,193 @@ func displaySuggestions(in string) error {
...
+
+	if len(tables) == 0 {
+		return nil, errors.New("nothing to dump")
+	}
+
+	return tables, nil
+}
+
+func runDumpCmd(db *genji.DB, table string) error {
+	switch table {
```
asdine, line 285: The code is the same for both cases, except that in one case we have 1 table and the other N tables. Let's try finding a way to simplify this

cmd/genji/shell/command.go
```diff
@@ -136,3 +141,193 @@ func displaySuggestions(in string) error {
 	fmt.Println()
 	return nil
 }
+
+func getDatabaseTransaction(db *genji.DB) (*database.Transaction, error) {
+	var dtx *database.Transaction
+	var err error
+	err = db.View(func(tx *genji.Tx) error {
```
asdine, line 148: This code does this:

- Use DB to create a transaction A
- Use transaction A to get DB back
- Use DB to create transaction B
- Close transaction A

Why not just use `db.Begin(false)`?

cmd/genji/shell/command.go
```diff
@@ -136,3 +141,193 @@ func displaySuggestions(in string) error {
...
+			if err != nil {
+				return err
+			}
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+
+		dtx, err := getDatabaseTransaction(db)
```
asdine, line 322: We don't need to create a transaction and commit it, we only need to display that statement. You can remove `getDatabaseTransaction` and `Commit` as it has no effect.

cmd/genji/shell/command.go
```diff
@@ -136,3 +141,193 @@ func displaySuggestions(in string) error {
...
+	}
+
+	return tables, nil
+}
+
+func runDumpCmd(db *genji.DB, table string) error {
+	switch table {
+	case "":
+		// Get all tables and ensure there is at least one table.
+		tables, err := getAllTables(db)
```
asdine, line 288: This loads all the tables in memory and then dump them, we should do them one by one instead.

asdine (changes requested):
I think this can be simplified much more:

- Open only one read only transaction for the whole command
- If we want to dump everything, list the table names and for each, call the dumpTable function with that each name
- If we want to dump only the table name, call that same function with that table name
- Rollback the transaction

cmd/genji/shell/command.go
```diff
@@ -136,3 +141,193 @@ func displaySuggestions(in string) error {
...
+	} else {
+		fmt.Printf("CREATE TABLE %s(", table)
+		for i, f := range fields {
+			fmt.Printf("%s %s", fields[i].Path, strings.ToUpper(fields[i].Type.String()))
+			if f.IsPrimaryKey {
+				fmt.Printf(" PRIMARY KEY")
+			}
+
+			if f.IsNotNull {
+				fmt.Printf(" IS NOT NULL")
```
asdine, line 193: This is not correct, this should be `NOT NULL`.
Also, please test the dump manually to see if everything is correct

cmd/genji/shell/command.go
tzzed: In fact `tx *genji.Tx` is sufficient.

cmd/genji/shell/command.go
tzzed: Well seen.

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
 	fmt.Println()
 	return nil
 }
+
+// dumpTable display the content of the given table as SQL statements.
```
asdine, line 144:
```suggestion
// dumpTable displays the content of the given table as SQL statements.
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+func dumpTable(db *genji.DB, t *database.Table) error {
+	// Get table info.
+	ti, err := t.Info()
+	if err != nil {
+		return err
+	}
+
+	fields := ti.FieldConstraints
+	if len(fields) == 0 {
+		fmt.Printf("CREATE TABLE %s;\n", t.Name())
```
asdine, line 154: We should use a [strings.Builder](https://golang.org/pkg/strings/#Builder) instead of printing chunks. If there is an error in the middle of printing a statement we might end up with weird output.
Or even a better, a [bytes.Buffer](https://golang.org/pkg/bytes/#Buffer) combined with [fmt.Fprintf](https://golang.org/pkg/fmt/#Fprintf)

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+	} else {
+		fmt.Printf("CREATE TABLE %s(", t.Name())
```
asdine, line 156:
```suggestion
		fmt.Printf("CREATE TABLE %s (", t.Name())
```

cmd/genji/shell/shell.go
```diff
@@ -260,6 +260,18 @@ func (sh *Shell) runCommand(in string) error {
...
 		}
 		return runIndexesCmd(db, cmd)
+	case ".dump":
+		db, err := sh.getDB()
+		if err != nil {
+			return err
+		}
+
+		t := ""
+		if len(cmd) == 2 {
```
asdine, line 270: What happens if I write `.dump foo bar baz`?

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+			fmt.Printf("INSERT INTO %s VALUES %s;\n", t.Name(), buf.String())
+			return nil
+		})
+
+	})
+}
+
+func runDumpCmd(db *genji.DB, table string) error {
+	fmt.Println("BEGIN TRANSACTION;")
+	defer fmt.Println("COMMIT;")
```
asdine, line 231: If there is an error in the middle of the dump, we will output `COMMIT` anyway, we should not do this. Let's print it when it's supposed to be printed instead.

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+			}
+
+			return dumpTable(db, t)
+		})
+	}
+
+	// Dump database content.
+	return db.View(func(tx *genji.Tx) error {
+		i := 0
+		return tx.TableInfoStoreIterate(func(table string, ti database.TableInfo) error {
```
asdine, line 248: There is a better way of doing this: Just use SQL :)

```sql
SELECT table_name FROM __genji_tables

File: cmd/genji/shell/command.go [hunk 14]
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+			if f.IsPrimaryKey {
+				fmt.Printf(" PRIMARY KEY")
+			}
+
+			if f.IsNotNull {
+				fmt.Printf(" NOT NULL")
+			}
+
+			// Don't print the last comma.
+			if i == (len(fields) - 1) {
```
asdine, line 168: Here is a more elegant solution: At the beginning of the loop, just check if i > 0, if so, print the comma.

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+	indexes, err := t.Indexes()
+	if err != nil {
+		return err
+	}
+
+	// Display indexes.
+	for _, index := range indexes {
+		u := ""
+		if index.Opts.Unique {
+			u = " UNIQUE"
```
asdine, line 189:
```suggestion
			u = "UNIQUE"
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+		}
+
+		info := fmt.Sprintf("CREATE%s INDEX %s ON %s (%s);", u, index.Opts.IndexName, index.Opts.TableName, index.Opts.Path)
```
asdine, line 192:
```suggestion
		info := fmt.Sprintf("CREATE %s INDEX %s ON %s (%s);", u, index.Opts.IndexName, index.Opts.TableName, index.Opts.Path)
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
 	fmt.Println()
 	return nil
 }
+
+// dumpTable display the content of the given table as SQL statements.
+func dumpTable(db *genji.DB, t *database.Table) error {
```
asdine, line 145: `dumpTable` is always called from within a transaction, we should pass the transaction instead of `db`

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+		if index.Opts.Unique {
+			u = " UNIQUE"
+		}
+
+		info := fmt.Sprintf("CREATE%s INDEX %s ON %s (%s);", u, index.Opts.IndexName, index.Opts.TableName, index.Opts.Path)
+		fmt.Println(info)
+	}
+
+	// Inserts section.
+	return db.View(func(tx *genji.Tx) error {
```
asdine, line 197: We are opening a transaction from within a transaction, we should simply use the one created outside of the call to `dumpTable`

cmd/genji/shell/command.go
```diff
@@ -136,3 +140,132 @@ func displaySuggestions(in string) error {
...
+		q := fmt.Sprintf("SELECT * FROM %s", t.Name())
+		res, err := tx.Query(q)
+		if err != nil {
+			return err
+		}
+		defer res.Close()
+
+		c, err := res.Count()
```
asdine, line 205: This will loop over all the documents unnecessarily. If the table is empty `res.Iterate` will simply return nil without calling the inner function, we don't need to have a special case here

cmd/genji/shell/shell.go
```diff
@@ -32,7 +32,7 @@ type Shell struct {
 	livePrefix string
 	multiLine  bool
 
-	history        []string
```
asdine, line 35: I think if you rebase with master this is already formatted correctly

document/json.go
```diff
@@ -25,7 +25,6 @@ func NewFromJSON(data []byte) (Document, error) {
 // MarshalJSON implements the json.Marshaler interface.
 func (v Value) MarshalJSON() ([]byte, error) {
 	var x interface{}
-
```
asdine, line 28: Could you explain this change?

asdine (changes requested):
This can be simplified even further 💪🏼

cmd/genji/shell/shell.go
tzzed: I will see.

cmd/genji/shell/command.go
tzzed: I was to defensive.

cmd/genji/shell/shell.go
tzzed: What do you want? Should I print the content of the three args ? And so change it to `runDumpCmd(db, cmd[1:]) `

cmd/genji/shell/command.go
tzzed: I put a white space too avoids this:
```SQL
...
foo.bar[1].hello BLOBPRIMARY KEY,
foo.a[1][2] TEXTNOT NULL,
...
```

cmd/genji/shell/shell.go
asdine: Yeah that's a good idea

cmd/genji/shell/command.go
asdine: This is about the index creation not about the table creation, look below you are running `fmt.Sprintf("CREATE INDEX")`

cmd/genji/shell/shell.go
tzzed: Ok What about the error handling. Let's admit `.dump foo bar baz` where `bar` doesn't exists?

cmd/genji/shell/command.go
tzzed: Yes I do it for the same reason. 
With `u = " UNIQUE"`
```SQL
CREATE UNIQUE INDEX idx ON test (foo[1]);
```

Without

```SQL
CREATE  INDEX idx_foo_a ON foo (a);
```

I get 2 spaces between `CREATE` and `INDEX`.

cmd/genji/shell/command.go
tzzed: If there is an error what kind of message? sqlite3 print `ROLLBACK; --due to an error` should we do it or don't print anything. 
```SQL
BEGIN TRANSACTION;
error: XXX 
```

cmd/genji/shell/command.go
asdine: Ok it makes sense 👍🏼

cmd/genji/shell/command.go
asdine: I think we should print the rollback yes, this might be streamed to another instance of Genji who will get statements one by one, we want to make sure that if something happens in the middle of the process, we keep a consistent state.

cmd/genji/shell/shell.go
asdine: Good question, I think we can ignore them and display BEGIN then COMMIT.

```sql
.dump tableThatDoesntExist
BEGIN TRANSACTION:
COMMIT;
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+				fmt.Println()
+			}
+			i++
+
+			if err := dumpTable(tx, t); err != nil {
+				fmt.Println("ROLLBACK;")
+				return err
+			}
+		case database.ErrTableNotFound: // If table doesn't exist we skip it.
+			break
```
asdine, line 278: `break` is ambiguous because of its meaning in other languages, only use it when you want to explicitly stop the iteration on a loop. If you want to skip this loop iteration, use `continue` instead, or if you want to do nothing, just write nothing.

```suggestion
			continue
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+		default:
+			fmt.Println("ROLLBACK;")
+			return err
+		}
+	}
+
+	// tables slice argument is not empty, all args tables has been displayed.
+	// If it is empty we should print the database content.
+	if argsTable {
```
asdine, line 287: This can be replaced by `len(tables) > 0`

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+	defer tx.Rollback()
+
+	fmt.Println("BEGIN TRANSACTION;")
+
+	// Dump the given table(s) content.
+	argsTable := false
+	i := 0
+	for _, table := range tables {
+		argsTable = true
+		t, err := tx.GetTable(table)
```
asdine, line 264: It seems that we are doing this work twice, once here and once below. How about passing the table name directly to `dumpTable` and let that function get the table and handle the error?

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
 	fmt.Println()
 	return nil
 }
+
+// dumpTable displays the content of the given table as SQL statements.
+func dumpTable(tx *genji.Tx, t *database.Table) error {
+	var buf bytes.Buffer
+
+	c := fmt.Sprintf("CREATE TABLE %s", t.Name())
+	buf.WriteString(c)
```
asdine, line 151:
```suggestion
	c := fmt.Fprintf(&buf, "CREATE TABLE %s", t.Name())
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+
+	ti, err := t.Info()
+	if err != nil {
+		return err
+	}
+
+	fcs := ti.FieldConstraints
+	hasField := false
```
asdine, line 159: Calling `len` is very cheap, there is no need to create temp booleans for that

```go
	if len(fcs) >0 {
	  buf.WriteString("(\n")
	}

        for ... {

         }

         if len(fcs) > 0 {
             buf.WriteString(")\n")
         }
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+		return err
+	}
+
+	for _, index := range indexes {
+		u := ""
+		if index.Opts.Unique {
+			u = " UNIQUE"
+		}
+
+		info := fmt.Sprintf("CREATE%s INDEX %s ON %s (%s);\n", u, index.Opts.IndexName, index.Opts.TableName,
```
asdine, line 209: There is no reason for using the buffer here, you can do this in one action with `fmt.Printf`

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+	}
+
+	q := fmt.Sprintf("SELECT * FROM %s", t.Name())
+	res, err := tx.Query(q)
+	if err != nil {
+		return err
+	}
+	defer res.Close()
+
+	// Inserts statement.
```
asdine, line 227:
```suggestion
	// Insert statements.
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+
+	ti, err := t.Info()
+	if err != nil {
+		return err
+	}
+
+	fcs := ti.FieldConstraints
+	hasField := false
+	for i, fc := range fcs {
+		// Fields constraints should be displaying between parenthesis.
```
asdine, line 161:
```suggestion
		// Fields constraints should be displayed between parenthesis.
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
...
+		if !hasField {
+			buf.WriteString("(\n")
+			hasField = true
+		}
+
+		// Don't put the last comma.
```
asdine, line 167:
```suggestion
		// Don't display the last comma.
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,193 @@ func displaySuggestions(in string) error {
 	fmt.Println()
 	return nil
 }
+
+// dumpTable displays the content of the given table as SQL statements.
+func dumpTable(tx *genji.Tx, t *database.Table) error {
```
asdine, line 147: We should receive an `io.Writer` and write to it, instead of writing directly to stdout:
- this will make this function almost ready to be used as a separate command
- we will be able to write tests for this, which we don't have at the moment

asdine (changes requested):
Thanks @tzzed ! It seems functionally correct but I have a few comments to improve the overall design.
Also, there are no tests, we should add some. I added a comment on how to do that by using an `io.Writer` instead of writing to `os.Stdout` directly

cmd/genji/shell/command.go
tzzed: I didn't dare to do it.

cmd/genji/shell/command.go
tzzed: You are right and `break` doesn't do what I want. Mistakenly wrote.

cmd/genji/shell/command.go
tzzed: I should use `buf` because the output can be of two kinds:
```SQL
CREATE TABLE test;

# OR

CREATE TABLE test ( a INTEGER);
```
The `fmt.Fprintf` should be done after the check of `fieldsContraints`.

cmd/genji/shell/command.go
tzzed: Right 👍

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,182 @@ func displaySuggestions(in string) error {
...
+		}
+
+		return nil
+	})
+	if err != nil {
+		_, err = fmt.Fprintln(w, "ROLLBACK;")
+		return err
+	}
+
+	_, err = fmt.Fprintln(w, "ROLLBACK;")
```
asdine, line 321:
```suggestion
	_, err = fmt.Fprintln(w, "COMMIT;")
```

cmd/genji/shell/command_test.go
```diff
@@ -81,3 +83,91 @@ func TestIndexesCmd(t *testing.T) {
...
+			return func(t *testing.T) {
+				db, err := genji.Open(":memory:")
+				require.NoError(t, err)
+				defer db.Close()
+
+				var bwant bytes.Buffer
+
+				tx := fmt.Sprintf("%s\n", `BEGIN TRANSACTION;`)
+				bwant.WriteString(tx)
+				ci := fmt.Sprintf("%s\n", `COMMIT;`)
```
asdine, line 112:
```suggestion
				tx := "BEGIN TRANSACTION;\n"
				bwant.WriteString(tx)
				ci := "COMMIT;\n"
```

asdine (changes requested):
Keep going, we're almost there 💪🏼

cmd/genji/shell/command_test.go
```diff
@@ -81,3 +83,91 @@ func TestIndexesCmd(t *testing.T) {
...
+
+			}
+		}
+
+		t.Run("No Index/"+tt.name, testFn(false, false))
+		t.Run("With Index/"+tt.name, testFn(true, false))
+		t.Run("With FieldsConstraints/"+tt.name, testFn(true, true))
+		t.Run("With FieldsConstraints/"+tt.name, testFn(true, true))
+	}
+
```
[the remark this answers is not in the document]
tzzed: I use it but I forgot to run it.

cmd/genji/shell/command.go
tzzed: Even if it is not `ErrTableNotFound` we do a `COMMIT;`

sql/query/insert_test.go
```diff
@@ -216,7 +216,7 @@ func TestInsertStmt(t *testing.T) {
 
 		for i, test := range tests {
 			t.Run(test.name, func(t *testing.T) {
-				q := fmt.Sprintf("CREATE TABLE test%d (a %s)", i, test.fieldConstraint)
+				q := fmt.Sprintf( "CREATE TABLE test%d (a %s)", i, test.fieldConstraint)
```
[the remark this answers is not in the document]
tzzed: Sorry

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,183 @@ func displaySuggestions(in string) error {
...
+		err = dumpTable(tx, table, w)
+		switch err {
+		case nil:
+			// Blank separation between tables.
+			if i > 0 {
+				if _, err := fmt.Fprintln(w, ""); err != nil {
+					return err
+				}
+			}
+
```
asdine, line 270:
```suggestion
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,183 @@ func displaySuggestions(in string) error {
...
+		case database.ErrTableNotFound: // If table doesn't exist we skip it.
+			continue
+		default:
+			_, err = fmt.Fprintln(w, "COMMIT;")
+			return err
+		}
+	}
+
+	// tables slice argument is not empty, all args tables has been displayed.
```
asdine, line 279:
```suggestion
	// tables slice argument is not empty, all tables have been dumped.
```

cmd/genji/shell/command.go
```diff
@@ -136,3 +142,183 @@ func displaySuggestions(in string) error {
...
+		var tableName string
+		if err := document.Scan(d, &tableName); err != nil {
+			return err
+		}
+
+		if err := dumpTable(tx, tableName, w); err != nil {
+			return err
+		}
+
+		return nil
```
asdine, line 315:
```suggestion
		return dumpTable(tx, tableName, w)
```

cmd/genji/shell/command.go
asdine: > This is about the index creation not about the table creation, look below you are running fmt.Sprintf("CREATE INDEX")

@tzzed

asdine (commented):
That looks good @tzzed ! There is still an unresolved comment and I added a few quick changes that would slightly improve the PR

asdine:
@tzzed I'll merge, I'll deal with the changes myself as I'm gonna work on the cli. Thank you!

89z:
https://github.com/genjidb/genji/issues/315

PR #205, opened 2020-09-27
Sell command can be ended by a semicolon #198

### LINKED ISSUE #198: Shell command can be ended by a semicolon
Usually in the shell queries should be ended by a semicolon `;`.
It would be convenient to be able to put a semicolon for commands.
```
genji>.indexes foo;
idx_foo_a on foo(a)
genji>
```

u5surf:
Fixes #198 

shell commands are to be allowed the suffix of semicolon.

### DIFF
--- a/cmd/genji/shell/shell.go
+++ b/cmd/genji/shell/shell.go
@@ -238,6 +238,7 @@ func (sh *Shell) executeInput(in string) error {
 }
 
 func (sh *Shell) runCommand(in string) error {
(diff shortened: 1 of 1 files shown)


asdine (changes requested):
Thanks @u5surf ! I think this can be simplified even further by trimming right before determining which command it is

asdine (approved):
Thank you @u5surf !

PR #209, opened 2020-09-28
Fix benchmarks for badgerengine

tie:
Running `go test -bench Scan` in `engine/badgerengine` panics.
See the commit message for the details.


asdine (approved):
Thank you @tie !

Issue #210, opened 2020-09-28
Using driver/sql hangs database and other oddities


tie:
I can reproduce mattn/go-sqlite3#204 with [minimal changes](https://gist.github.com/tie/88322c6f880f8515b3b5ac64b2124ba4/revisions#diff-53618cae32e23ae47e1c35f10da92654).

<details>
<summary>main1.go</summary>

```
package main

import (
	"context"
	"database/sql"

	_ "github.com/genjidb/genji/sql/driver"
)

func main() {
	ctx := context.Background()

	db, err := sql.Open("genji", ":memory:")
	if err != nil {
		panic(err)
	}
	c1, err := db.Conn(ctx)
	if err != nil {
		panic(err)
	}
	c2, err := db.Conn(ctx)
	if err != nil {
		panic(err)
	}

	if _, err := c1.ExecContext(ctx, "CREATE TABLE m"); err != nil {
		panic(err)
	}
	if _, err := c2.ExecContext(ctx, "INSERT INTO m (k, v) VALUES ('k', 'v')"); err != nil {
		panic(err) // table not found
	}
	panic("unreachable")
}
```
</details>

<details>
<summary>main2.go</summary>

```
package main

import (
	"context"
	"database/sql"

	_ "github.com/genjidb/genji/sql/driver"
)

func main() {
	ctx := context.Background()

	db, err := sql.Open("genji", "test.db")
	if err != nil {
		panic(err)
	}
	db.Conn(ctx)
	db.Conn(ctx)
	panic("unreachable")
}
```
</details>

I assume SQLite releases db locks when they are not needed. But this issue gets worse with Genji since BoltDB blocks on db lock (unless a timeout was specified, which is *not* currently possible with `sql/driver`), and BadgerDB I believe just returns an error. That is, things go terribly wrong when `sql.DB` attempts to open a new `sql.Conn`.

As a temporary workaround it’s possible to call `SetMaxOpenConns(1)` on `*sql.DB` from the user’s code. This should work just fine for in-memory and bolt engines, but afaik badger supports concurrent transactions whereas single `driver.Conn` can’t be used concurrently.

I’ve experimented a bit with possible APIs to fix those issues, and I think the best option would be to fix this use case on the Go side by allowing `driver.Connector` to implement optional `io.Closer` interface and close it in `sql.DB.Close`. Then `driver.Connector` would be able to hold `genji.DB` object and release the resources when `sql.DB` is no longer needed.

[...]

PR #213, opened 2020-10-01
Implement driver.DriverContext and driver.Connector interfaces in sql/driver

tie:
This PR partially addresses issue #210 by implementing the approach that relies either on Go support (see golang/go#41710) or runtime.SetFinalizer to close the underlying database. Note that the latter is non-deterministic and we can’t handle an error.


asdine (approved):
Nice! I really hope https://github.com/golang/go/pull/41710 gets accepted eventually 👍🏼

PR #215, opened 2020-10-01
Respect NO_COLOR and NO_HISTORY environment variables

tie:
This PR allows disabling colored prompt by setting NO_COLOR environment variable. It also allows disabling history dumping and loading (that is, only I/O) in a similar fashion.

See https://no-color.org


asdine (approved):
Good idea! Thanks!

PR #216, opened 2020-10-01
Be more forgiving when scanning null values in StructScan

### LINKED ISSUE #211: Should be able to cast null type into array
the array should be nil

tie:
Fixes #211


asdine (approved):
Thanks @tie !

PR #218, opened 2020-10-02
Automatically create a database and a table with genji insert option '-a'

### LINKED ISSUE #217: CLI: Automatically create a database and a table with genji insert
Currently `genji insert` requires specifying the database name and the table name.

```bash
# first create a table
echo 'CREATE TABLE foo' | genji my.db

# then insert data from the command line

# like this
genji insert --db my.db -t foo '{"a": 1}'

# or like this
curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.db -t foo
```

It would be great if we could quickly store data without having to create that table beforehand. `insert` would create a database and a table automatically.

### New option: -a

```
-a, --auto
Automatically creates a database and a table whose name is equal to "data_" followed by the current unix timestamp.
It can be combined with --db to select an existing database but automatically create the table.
```

Example:

```bash
$ curl https://api.github.com/repos/genjidb/genji/issues | genji insert -a
$ ls
data-1601619182.db
$ genji data_1601619182.db
genji> .tables
data_1601619182
$ curl https://api.github.com/repos/genjidb/genji/issues | genji insert -a
$ ls
data_1601619182.db
data_1601619681.db
$ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db data_1601619681.db -a
$ ls
data_1601619182.db
data_1601619681.db
$ genji data_1601619681.db
genji> .tables
data_1601619681
data_1601620302
```

I'm open to suggestions concerning the option name, database and tables names.
I don't think this feature should worry too much about filename uniqueness and ordering, if this is important, users can use `--db` and `-t`.

Concerning the option name, `-c, --create` could also work.

tdakkota:
Fixes #217.

### DIFF
--- a/cmd/genji/insert.go
+++ b/cmd/genji/insert.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/dgraph-io/badger/v2"
 	"github.com/genjidb/genji"
@@ -108,9 +110,19 @@ func executeInsertCommand(db *genji.DB, table string, r io.Reader) error {
 	return nil
 }
 
-func runInsertCommand(e, DBPath, table string, args []string) error {
+func runInsertCommand(e, DBPath, table string, auto bool, args []string) error {
 	var ng engine.Engine
 	var err error
+
+	ts := time.Now().Unix()
+	if DBPath == "" && auto {
+		DBPath = "data_" + strconv.FormatInt(ts, 10) + ".db"
+	}
+	var createTable = false
+	if table == "" && auto {
+		table = "data_" + strconv.FormatInt(ts, 10)
+		createTable = true
+	}
 
 	switch e {
 	case "bolt":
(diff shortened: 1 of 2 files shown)


cmd/genji/insert.go
```diff
@@ -108,10 +110,20 @@ func executeInsertCommand(db *genji.DB, table string, r io.Reader) error {
 	return nil
 }
 
-func runInsertCommand(e, DBPath, table string, args []string) error {
+func runInsertCommand(e, DBPath, table string, auto bool, args []string) error {
```
asdine, line 113:
```suggestion
func runInsertCommand(e, dbPath, table string, auto bool, args []string) error {
```

cmd/genji/insert.go
```diff
@@ -108,10 +110,20 @@ func executeInsertCommand(db *genji.DB, table string, r io.Reader) error {
...
 	var ng engine.Engine
 	var err error
 
+	ts := time.Now().Unix()
+	if DBPath == "" && auto {
+		DBPath = "data_" + strconv.FormatInt(ts, 10) + ".db"
```
asdine, line 119: This only works with Bolt, because Badger creates a directory instead of a file.
We should take that into account when generating the path.

I think we should simply name the directory the same way, just without the ".db" suffix.

asdine (changes requested):
Thanks @tdakkota ! I have just one comment about badger 👍🏼

asdine (approved):
Thank you @tdakkota !

PR #223, opened 2020-10-04
Fix parentheses parsing

### LINKED ISSUE #207: IN doesn't work with parentheses
```
genji> select 1 IN (1, 2, 3);
found ,, expected ) at line 1, char 15
genji>
```

```
sqlite> select 1 IN (1, 2, 3);
1
sqlite>
```

Other example:

```sql
SELECT * FROM foo WHERE a IN (1, 2, 3)
```

This is a valid SQL expression, but currently, we only support this:

```sql
SELECT * FROM foo WHERE a IN [1, 2, 3]
```

This might be more complex than just parsing parentheses when we are in an IN operator.

I think we need to parse parentheses in two different ways:

### Single expression: `(expr)`

Example:

```sql
SELECT (1);
-> 1
SELECT (1 + 1);
-> 2
```

This is already supported, the parser [returns](https://github.com/genjidb/genji/blob/9da693665fe93d3fcfdffe3312a48712fc40023f/sql/parser/expr.go#L204-L212) an [expr.Parentheses](https://github.com/genjidb/genji/blob/9da693665fe93d3fcfdffe3312a48712fc40023f/sql/query/expr/expr.go#L127:6) expression.

### List of expressions: `(expr, expr, ...)`

Example:

```sql
SELECT (1, 2);
-> [1, 2]
SELECT (1 + 1, 3 > 5);
-> [2, false]
```

If we have more than one element in that list of parentheses, this must be parsed as an [expr.LiteralExprList](https://github.com/genjidb/genji/blob/9da693665fe93d3fcfdffe3312a48712fc40023f/sql/query/expr/literal.go#L75). 
There is already a function that can parse this kind of expression: https://github.com/genjidb/genji/blob/9da693665fe93d3fcfdffe3312a48712fc40023f/sql/parser/expr.go#L418:18

tdakkota:
Fixes #207.

### DIFF
--- a/sql/parser/expr.go
+++ b/sql/parser/expr.go
@@ -212,7 +212,7 @@ func (p *Parser) parseUnaryExpr() (expr.Expr, error) {
 		case scanner.RPAREN:
 			return expr.Parentheses{E: e}, nil
 		case scanner.COMMA:
-			exprList, err := p.parseListUntil(scanner.RPAREN)
+			exprList, err := p.parseExprListUntil(scanner.RPAREN)
 			if err != nil {
 				return nil, err
 			}
@@ -428,7 +428,7 @@ LOOP:
 	return vPath, nil
 }
 
-func (p *Parser) parseListUntil(rightToken scanner.Token) (expr.LiteralExprList, error) {
+func (p *Parser) parseExprListUntil(rightToken scanner.Token) (expr.LiteralExprList, error) {
 	var exprList expr.LiteralExprList
 	var expr expr.Expr
 	var err error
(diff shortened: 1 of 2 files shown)


asdine:
Thanks @tdakkota! I think the issue wasn't clear enough and misled you, I'm sorry about that. What we are trying to fix is the behavior of parentheses in general, not only in a `SELECT <constant IN expr>` setup.

Let's continue the discussion in #207

sql/parser/expr.go
```diff
@@ -415,12 +428,7 @@ LOOP:
 	return vPath, nil
 }
 
-func (p *Parser) parseExprList(leftToken, rightToken scanner.Token) (expr.LiteralExprList, error) {
-	// Parse ( or [ token.
-	if tok, pos, lit := p.ScanIgnoreWhitespace(); tok != leftToken {
-		return nil, newParseError(scanner.Tokstr(tok, lit), []string{leftToken.String()}, pos)
-	}
-
+func (p *Parser) parseListUntil(rightToken scanner.Token) (expr.LiteralExprList, error) {
```
asdine, line 431: Good idea!
I'd like to be pedantic here and rename this to `parseExprListUntil` for consistency

asdine (commented):
Thanks @tdakkota!

asdine (approved):

PR #226, opened 2020-10-05
Expose __genji_indexes table.

### LINKED ISSUE #214: Expose the __genji_indexes table
When we get Indexes of a table by using the following `func (t *Table) Indexes() (map[string]Index, error) `  that called by the following command, the order is randomized.
```
genji>.indexes test
idx_b on test(b)
idx_c on test(c)
idx_a on test(a)
genji> .indexes test;
idx_a on test(a)
idx_b on test(b)
idx_c on test(c)
```
expected behaviour:
```
genji> .indexes test;
idx_a on test(a)
idx_b on test(b)
idx_c on test(c)
genji> .indexes test;
idx_a on test(a)
idx_b on test(b)
idx_c on test(c)
```

tzzed:
This fixes #214.
I added `t.tableInfos[indexStoreName]` and fixed the `.indexes` command.

### DIFF
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -53,30 +53,25 @@ func runTablesCmd(db *genji.DB, cmd []string) error {
 // displayTableIndex prints all indexes that the given table contains.
 func displayTableIndex(db *genji.DB, tableName string) error {
 	ctx := context.Background()
-	res, err := db.Query(ctx, "SELECT index_name FROM __genji_indexes")
+	res, err := db.Query(ctx, "SELECT * FROM __genji_indexes")
 	if err != nil {
 		return err
 	}
 	defer res.Close()
 
 	return res.Iterate(func(d document.Document) error {
-		var index string
-		if err = document.Scan(d, &index); err != nil {
+		var index database.IndexConfig
+		err := index.ScanDocument(d)
+		if err != nil {
+			return err
+		}
+		if index.TableName != tableName {
 			return nil
 		}
 
-		return db.View(func(tx *genji.Tx) error {
-			idx, err := tx.GetIndex(index)
-			if err != nil {
-				return err
-			}
+		fmt.Printf("%s ON %s (%v)\n", index.IndexName, index.TableName, index.Path)
 
-			if idx.Opts.TableName == tableName {
-				fmt.Printf("%s on %s(%s)\n", idx.Opts.IndexName, idx.Opts.TableName, idx.Opts.Path)
-			}
-
-			return nil
-		})
+		return nil
 	})
 }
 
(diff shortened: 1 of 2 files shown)


cmd/genji/shell/command.go
```diff
@@ -52,43 +52,58 @@ func runTablesCmd(db *genji.DB, cmd []string) error {
 
 // displayTableIndex prints all indexes that the given table contains.
 func displayTableIndex(db *genji.DB, tableName string) error {
-	err := db.View(func(tx *genji.Tx) error {
-		t, err := tx.GetTable(tableName)
-		if err != nil {
-			return err
-		}
+	ctx := context.Background()
+	res, err := db.Query(ctx, "SELECT index_name FROM __genji_indexes")
```
asdine, line 56: By exposing the __genji_indexes table, you don't need to use `t.GetIndex` anymore, You know have access to all of these fields: https://github.com/genjidb/genji/blob/9da693665fe93d3fcfdffe3312a48712fc40023f/database/config.go#L377:L388

asdine (changes requested):
Thanks @tzzed ! Exposing the __genji_indexes table means you don't need to access the `database` package anymore. If you have trouble understanding what fields you can use, just run this on the CLI:

```sql
SELECT * FROM __genji_indexes;
```

cmd/genji/shell/command.go
tzzed: Thank you !

cmd/genji/shell/command.go
```diff
@@ -52,43 +52,50 @@ func runTablesCmd(db *genji.DB, cmd []string) error {
 
 // displayTableIndex prints all indexes that the given table contains.
 func displayTableIndex(db *genji.DB, tableName string) error {
-	err := db.View(func(tx *genji.Tx) error {
-		t, err := tx.GetTable(tableName)
-		if err != nil {
-			return err
-		}
+	ctx := context.Background()
+	res, err := db.Query(ctx, "SELECT * FROM __genji_indexes")
```
asdine, line 56: Why not use SQL for the filtering as well (i.e. `WHERE`)?

tzzed: `db.QueryDocument` I didn't know this function. 
About SQL it because sometimes I think weirdly. Change it.

cmd/genji/shell/command.go
```diff
@@ -52,43 +52,54 @@ func runTablesCmd(db *genji.DB, cmd []string) error {
 
 // displayTableIndex prints all indexes that the given table contains.
 func displayTableIndex(db *genji.DB, tableName string) error {
-	err := db.View(func(tx *genji.Tx) error {
-		t, err := tx.GetTable(tableName)
+	ctx := context.Background()
+
+	q := fmt.Sprintf("SELECT * FROM __genji_indexes WHERE table_name = '%s'", tableName)
```
asdine, line 57: For your culture, you should **never** do this in a user facing application. This allows anyone to run [sql injections](https://www.w3schools.com/sql/sql_injection.asp)

Here for example, let's imagine that this function is called by a command or a web app and that `tableName` is passed by a user. Someone could write the following:

```
'; DROP TABLE users; SELECT '1
```

This would generate this:

```sql
SELECT * FROM __genji_indexes WHERE table_name = ''; DROP TABLE users; SELECT '1'
```

This is a valid sql query, which would destroy the users table.

To avoid that, always use query parameters

`res, err := db.Query(ctx, "SELECT * FROM __genji_indexes WHERE table_name = ?", tableName)`

cmd/genji/shell/command.go
```diff
@@ -52,43 +52,54 @@ func runTablesCmd(db *genji.DB, cmd []string) error {
...
+	return res.Iterate(func(d document.Document) error {
+		var index database.IndexConfig
+		err := index.ScanDocument(d)
 		if err != nil {
 			return err
 		}
 
-		indexes, err := t.Indexes()
-		if err != nil {
+		if err := index.ScanDocument(d); err != nil {
```
asdine, line 71: We are running `ScanDocument` twice here

asdine (changes requested):

cmd/genji/shell/command.go
tzzed: Thank you Asdine.

asdine (approved):
Thanks @tzzed !

PR #228, opened 2020-10-06
Add AVG() aggregate function, allow to add custom via API 

### LINKED ISSUE #227: SQL: add more SQL functions, allow to add custom via API
For now all functions stored in global map, but I think it should be more customizable. 
I think it can be given to the `parser.Parser` type as argument.
https://github.com/genjidb/genji/blob/9da693665fe93d3fcfdffe3312a48712fc40023f/sql/query/expr/function.go#L12-L43

Also it would be great to support by default widely used SQL functions like `avg()`/ `rank()`

tdakkota:
Fixes #227.

### DIFF
--- a/sql/parser/parser.go
+++ b/sql/parser/parser.go
@@ -24,12 +24,16 @@ type Parser struct {
 
 // NewParser returns a new instance of Parser.
 func NewParser(r io.Reader) *Parser {
-	return &Parser{s: scanner.NewBufScanner(r), functions: expr.NewFunctions()}
+	return NewParserWithOptions(r, nil)
 }
 
-func (p *Parser) WithFunctions(functions expr.Functions) *Parser {
-	p.functions = functions
-	return p
+// NewParserWithOptions returns a new instance of Parser using given Options.
+func NewParserWithOptions(r io.Reader, opts *Options) *Parser {
+	if opts == nil {
+		opts = defaultOptions()
+	}
+
+	return &Parser{s: scanner.NewBufScanner(r), functions: opts.Functions}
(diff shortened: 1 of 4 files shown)


sql/parser/parser.go
```diff
@@ -19,11 +19,17 @@ type Parser struct {
...
 }
 
 // NewParser returns a new instance of Parser.
 func NewParser(r io.Reader) *Parser {
-	return &Parser{s: scanner.NewBufScanner(r)}
+	return &Parser{s: scanner.NewBufScanner(r), functions: expr.NewFunctions()}
+}
+
+func (p *Parser) WithFunctions(functions expr.Functions) *Parser {
+	p.functions = functions
```
tie, line 31: I think `WithFunctions` should return a shallow copy.

```go
p2 := *p
p2.functions = functions
return &p2
```

asdine: I think we shouldn't use the `With*` pattern here because as @tie suggested it's better to return shallow copies when doing so, which would force us to duplicate buffers, perhaps reset readers, etc.

What we want is to create a parser with some specific functions. How about we just make the constructor more flexible?

```go
func NewParser(r io.Reader, opts *Options) *Parser {
 ...
}

type Options struct {
  Functions expr.Functions
  // in the future we could add more options here
}
```

`nil` options means using defaults. We are already doing so in other packages

asdine (changes requested):
Thanks @tdakkota ! The functions management is definitely more clean now!

asdine (approved):
👍🏼

PR #229, opened 2020-10-06
Types Aliases

### LINKED ISSUE #183: Add type aliases
Supported types in Genji are the following:
- `INTEGER`
- `DOUBLE`
- `BOOL`
- `BLOB`
- `TEXT`
- `ARRAY`
- `DOCUMENT`

Add type aliases allowing to refer to any of these types using this list:

Types referring to `INTEGER`:
- `INT`
- `TINYINT`
- `SMALLINT`
- `MEDIUMINT`
- `BIGINT`
- `INT2`
- `INT8`

Types referring to `DOUBLE`:
- `DOUBLE PRECISION`
- `REAL`

Types referring to `TEXT`:
- `CHARACTER(n)`
- `VARCHAR(n)`

Any of these types, when parsed, must actually refer to the corresponding concrete type.

tzzed:
Fixes #183 

Add all Aliases types and tests for this types in different queries.


sql/scanner/token.go
```diff
@@ -221,14 +236,25 @@ var tokens = [...]string{
...
 
-	TYPEARRAY:    "ARRAY",
-	TYPEBLOB:     "BLOB",
-	TYPEBOOL:     "BOOL",
-	TYPEBYTES:    "BYTES",
-	TYPEDOCUMENT: "DOCUMENT",
-	TYPEDOUBLE:   "DOUBLE",
-	TYPEINTEGER:  "INTEGER",
-	TYPETEXT:     "TEXT",
+	TYPEARRAY:       "ARRAY",
```
asdine, line 239: We should remove the `TYPE` prefix for all the types and order them in alphabetical order like the rest of the keywords

sql/scanner/token.go
```diff
@@ -124,6 +124,21 @@ const (
...
+	// Aliases
+	INT
+	TINYINT
+	SMALLINT
+	MEDIUMINT
+	BIGINT
+	INT2
+	INT8
+
+	DOUBLEPRECISION
```
asdine, line 136: Just `PRECISION` should be fine

sql/parser/expr.go
```diff
@@ -286,11 +286,27 @@ func (p *Parser) parseType() document.ValueType {
...
 		return document.DoubleValue
-	case scanner.TYPEINTEGER:
+	case scanner.TYPEINTEGER, scanner.INT, scanner.INT2, scanner.INT8, scanner.TINYINT,
+		 scanner.BIGINT, scanner.MEDIUMINT, scanner.SMALLINT:
 		return document.IntegerValue
 	case scanner.TYPETEXT:
+		return document.TextValue
+	case scanner.VARCHAR, scanner.CHARACTER:
+		// The value between parentheses is not used.
+		_, err := p.parseExprList(scanner.LPAREN, scanner.RPAREN)
```
asdine, line 305: This would make the following valid:

```SQL
VARCHAR(1 IN [1, 2, 3] AND foo > 4)
```

I understand the trick but parsers should be very strict.
We should expect one left parenthesis, an integer, and a closing parenthesis.

sql/parser/expr.go
```diff
@@ -286,11 +286,27 @@ func (p *Parser) parseType() document.ValueType {
...
+		if err != nil {
+			return 0
```
asdine, line 307: Now that we have a case that can fail, we should change the signature of this function and return an error. In this case a `ParseError`

cmd/genji/shell/command_test.go
```diff
@@ -119,6 +119,7 @@ func TestRunDumpCmd(t *testing.T) {
 				if withConstraints {
 					q := fmt.Sprintf("CREATE TABLE test (\n  a %s\n);\n", tt.fieldConstraint)
 					err := db.Exec(ctx, q)
+
```
asdine, line 122: 👀

asdine (changes requested):
Thanks @tzzed ! We still need to normalize the const names and change a few things but it's going to the right direction

sql/scanner/token.go
tzzed: I didn't dare...

sql/parser/expr.go
```diff
@@ -559,3 +575,32 @@ func (p *Parser) parseCastExpression() (expr.Expr, error) {
...
+	_, _, err := p.ParseExpr()
+	if err != nil {
+		return nil, err
+	}
+
+	if tok, pos, lit := p.ScanIgnoreWhitespace(); tok != scanner.RPAREN {
+		return nil, newParseError(scanner.Tokstr(tok, lit), []string{")"}, pos)
+	}
+
+	return nil, nil
```
asdine, line 605: This is not necessary, isn't it?

sql/scanner/token.go
```diff
@@ -115,14 +115,27 @@ const (
...
-	TYPEDOUBLE
-	TYPEINTEGER
-	TYPETEXT
+	ARRAY
+	BLOB
+	BOOL
+	BYTES
+	DOCUMENT
+	DOUBLE
+	INTLITERAL // literal INTEGER
```
asdine, line 124: Sorry, now I remember why there was this `TYPE` prefix, it is to distinguish a type keyword and values.
We should use actually the `TYPE`  prefix for all the types

asdine: Concerning the `PRECISION` keyword, it should be in the same group as the other normal keywords above, ordered alphabetically.

sql/parser/expr.go
asdine: If we are not going to return the value inside (which is not an expression by the way but an integer), we might as well remove this function and use the 3 ifs above directly in the switch l318

asdine (changes requested):

sql/parser/expr.go
```diff
@@ -545,7 +568,7 @@ func (p *Parser) parseCastExpression() (expr.Expr, error) {
 	}
 
 	// Parse require typename.
-	tp := p.parseType()
+	tp, err := p.parseType()
```
asdine, line 571: We should handle this error

asdine:
@tzzed I think there was a rebase or merge error, the commits seem wrong. Could you take a look?

tzzed:
> @tzzed I think there was a rebase or merge error, the commits seem wrong. Could you take a look?

seems good now.

sql/scanner/scanner_test.go
```diff
@@ -159,6 +159,7 @@ func TestScanner_Scan(t *testing.T) {
 		{s: "DOUBLE", tok: scanner.TYPEDOUBLE, raw: `DOUBLE`},
 		{s: "INTEGER", tok: scanner.TYPEINTEGER, raw: `INTEGER`},
 		{s: "TEXT", tok: scanner.TYPETEXT, raw: `TEXT`},
+
```
asdine, line 162: Please avoid unnecessary changes

```suggestion

```

sql/scanner/token.go
```diff
@@ -221,14 +234,25 @@ var tokens = [...]string{
...
+	TYPEINT2:      "INT2",
+	TYPEINT8:      "INT8",
+	TYPEINTEGER:   "INTEGER",
+	TYPEMEDIUMINT: "MEDIUMINT",
+	TYPESMALLINT:  "SMALLINT",
+	TYPETEXT:      "TEXT",
+	TYPETINYINT:   "TINYINT",
+	TYPEREAL:      "REAL",
+	TYPEVARCHAR:   "VARCHAR",
+
```
asdine, line 255:
```suggestion
```

sql/query/create_test.go
```diff
@@ -117,6 +117,46 @@ func TestCreateTable(t *testing.T) {
...
+					{Path: parsePath(t, "b"), Type: document.IntegerValue},
+					{Path: parsePath(t, "m"), Type: document.IntegerValue},
+					{Path: parsePath(t, "eight"), Type: document.IntegerValue},
+					{Path: parsePath(t, "ii"), Type: document.IntegerValue},
+					{Path: parsePath(t, "c"), Type: document.TextValue},
+				}, info.FieldConstraints)
+				return nil
+			})
+			require.NoError(t, err)
+
```
asdine, line 157:
```suggestion
```

sql/parser/expr.go
```diff
@@ -558,4 +585,4 @@ func (p *Parser) parseCastExpression() (expr.Expr, error) {
 	}
 
 	return expr.CastFunc{Expr: e, CastAs: tp}, nil
-}
+}
```
asdine, line 588: This one as well

cmd/genji/shell/command.go
```diff
@@ -104,7 +104,6 @@ func displayAllIndexes(db *genji.DB) error {
 
 		return nil
 	})
-
```
asdine, line 107: 👀

asdine (changes requested):
It is functionally correct, however, there are too many unnecessary changes such as new lines added and removed

PR #230, opened 2020-10-07
Added replace rule for badgerengine in cmd/genji module

tie:
Replace in `go.mod` works only for the main module.


asdine (approved):
Thanks!

Issue #206, opened 2020-09-27
Add support for context.Context


asdine:
Currently, while there is a way of stopping `SELECT` queries by simply stopping the iteration on the stream, there is no way of halting `INSERT`, `UPDATES` and `DELETES`. When we'll add subqueries, it will be even more important.

We should change the api to pass `context.Context` everywhere it is necessary.

Issue #58, opened 2020-02-17
Support for FillPercent/bbolt


jscoys:
Hello

I’m trying to find a way to specify the FillPercent value on the bbolt Engine without success.

The default value is 0.5, would like to put a 0.9. Do you know how to proceed through Genji when creating the db/bucket?

Thx

jscoys:
@asdine can you help me please? Thx

asdine:
There is currently no way to customize the way the Bolt engine is running, besides the bolt options.

```go
db, err := genji.New(boltengine.New("my.db", 0600, &bolt.Options{...}))
```

`FillPercent` must be set up manually on the required bucket on every transaction, which is not possible at this moment. Also, it would also mean that it should be table-specific, which is even more complicated. 

Perhaps it would be interesting to be able to customize the engine further, with something like this:

```go
db, err := genji.New(boltengine.New("my.db", &boltengine.Options{
  FillPercent: 0.9,
  Mode: 0600,
  Options: &bolt.Options{},
}))
```

but it means having a FillPercent value that would be the same for all buckets.

Could you describe why you need to have access to the FillPercent feature?

jscoys:
Well first thx for your answer!

It’s the first time I’m using bbolt and I’m adding a basic struct with 10 fields (mix of string/date/int) in a loop adding a total of 12,000 records and I end up with a DB size of 1 gigs. It seems to me that it’s big for the data I’m putting there...

I red on some forums that of the db is reaching a size it’s doubling the size with « blank spaces ». It’s allowing the dB to grow faster (scale better), but for the space it’s not optimal at all... and it seems this theory is good because I see the dB size growing like 32, then 64, 128, 256, 512 and then jumping directly to 1024 mbytes. On these forums people are saying that we can change the FillPercent per bucket, and putting a 0.9 value would solve the issue. That’s why I’m asking.

Hum won’t be proud of it but can I change something directly in the code of the package locally to make it possible? Like a hack I would say?

jscoys:
Oh just red again your second option it’s not an issue to have the same FillPercent for all buckets! So I can create the dB with a Genji.New()? I will give it a try as soon as I can.

asdine:
Closing this as Bolt is no longer supported

PR #231, opened 2020-10-07
Add support for context.Context

### LINKED ISSUE #224: Add support for context.Context in engines
PR #221 added support for passing context to high-level APIs. We should also pass them to the engine implementations. E.g. it’d be possible to pass context.Context to `tikvengine` (though currently there is no such engine, but the API maps well to the Genji engine definitions).

See also: #206

tie:
This PR adds `context.Context` support in Genji, starting with the `engine` package and then fixing compile errors everywhere. It doesn’t introduce any cancellation behavior though—that should be a separate, smaller, PR.

- `engine.Iterator` usage now requires a `it.Err()` check after the loop.
  ```
  it := st.Iterator(engine.IteratorOptions{})
  defer it.Close()

  for it.Seek(ctx, nil); it.Valid(); it.Next(ctx) {
  	…
  }
  if err := it.Err(); err != nil {
  	…
  }
  ```

  Notice that `Seek` and `Next` now accept a `context.Context` parameter. If an error occurs, `Valid` returns false.

- `database.Table` no longer directly implements `document.Iterator` since iteration may be I/O bound.

  ```
  // Before
  func (*Table) Iterate(func(d document.Document) error) error
  // After
  func (*Table) Iterator(context.Context) document.IteratorFunc
  func (*Table) Iterate(context.Context, func(d document.Document) error) error
  ```

Closes #224 and #206

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ func main() {
     defer res.Close()
 
     // Iterate over the results
-    err = res.Iterate(context.Background(), func(d document.Document) error {
+    err = res.Iterate(func(d document.Document) error {
         // When querying an explicit list of fields, you can use the Scan function to scan them
         // in order. Note that the types don't have to match exactly the types stored in the table
         // as long as they are compatible.
--- a/README.md
+++ b/README.md
@@ -73,10 +73,8 @@ import (
 )
 
 func main() {
-    ctx := context.Background()
-
     // Create a database instance, here we'll store everything on-disk using the BoltDB engine
-    db, err := genji.Open(ctx, "my.db")
+    db, err := genji.Open(context.Background(), "my.db")
     if err != nil {
         log.Fatal(err)
     }
@@ -84,16 +82,16 @@ func main() {
     defer db.Close()
 
     // Create a table. Schemas are optional, you don't need to specify one if not needed
-    err = db.Exec(ctx, "CREATE TABLE user")
+    err = db.Exec("CREATE TABLE user")
 
     // Create an index
-    err = db.Exec(ctx, "CREATE INDEX idx_user_name ON test (name)")
+    err = db.Exec("CREATE INDEX idx_user_name ON test (name)")
 
     // Insert some data
-    err = db.Exec(ctx, "INSERT INTO user (id, name, age) VALUES (?, ?, ?)", 10, "Foo1", 15)
+    err = db.Exec("INSERT INTO user (id, name, age) VALUES (?, ?, ?)", 10, "Foo1", 15)
 
     // Supported values can go from simple integers to richer data types like lists or documents
-    err = db.Exec(ctx, `
+    err = db.Exec(`
     INSERT INTO user (id, name, age, address, friends)
     VALUES (
         11,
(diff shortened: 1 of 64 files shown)


asdine:
I know the PR is still a draft but since it contains pretty heavy changes I'd like to comment now to prevent you from doing unnecessary work if agree on a different path.

Basically, I would like to avoid context pollution where it is not necessary.

IMO the benefit of context in the case of Genji is essentially for canceling anything that relies on IO, even indirectly.

Some APIs are designed to be one time in-memory procedures. For these, handling cancelation from within the function is the same as handling it from outside, so passing the context is not necessary. I'm thinking about the planner package for example, which takes a finite, not so big AST. I think it's ok to let it finish before checking if a context is canceled.

Even on IO related APIs, I believe we can avoid context pollution by having some indirection. For example, instead of passing the context to any engine/store method, we could pass it only to the `Begin` method and leave the implementation deal with the context for that particular transaction.
From a high level point of view, we want to be able to cancel a transaction, we are not interested in the lifecycle of each particular call to Get or Put.

Concerning the document.Iterator type, it is returned as a result of `db.Query`, which itself takes a context.
I would be in favor of not passing the context to `Iterate` and have iterate return `ctx.Err()` if ctx is done.
By itself, `document.Iterator` lets the caller control the flow of the stream. If a user wants to stop, it simply needs to stop iterating.

tie:
Thanks for the input!

I agree on the context pollution issue, but I’d rather avoid indirection as it doesn’t add much value since you either already have `context.Context` instance e.g. from HTTP request, or can easily instantiate `context.Background()`. Explicit contexts also allow passing some engine-specific per-operation values via context, which could be useful in cases like #58.

In Genji we can have the root `genji` package hide the explicit context passing using `WithContext` pattern, e.g.

```
var db *genji.DB
var tx *genji.Tx

db, _ := genji.Open(ctx, …)          // sets default context to ctx
tx, _ = db.Begin(…)                  // uses default context
db = db.WithContext(ctx)             // overrides default context
tx, _ = db.WithContext(ctx).Begin(…) // overrides per function call
tx.Exec(…)                           // uses context inherited on Begin
tx = tx.WithContext(ctx)
tx.WithContext(ctx).Exec(…)
```

Here, `WithContext` returns a shallow copy with a new embedded context. It’s cheap since `DB` and `Tx` just wrap the corresponding types in `database` package, and I’d even expect the compiler to optimize `WithContext` call if it’s possible to allocate a copy on stack.

The reason why `document.Iterator.Iterate` and `Optimize` accept context is that the former is implemented by `database.Table` (that is, the main concern here is that iteration may be I/O bound), and the later has `UseIndexBasedOnSelectionNodeRule` which calls I/O bound APIs.

https://github.com/genjidb/genji/blob/950009f414495518d5ca510dd604ec81385dfa7a/sql/planner/optimizer.go#L281

I believe Go’s team motivation for FS package is to be backwards and API-compatible with `io` and `os` interfaces.

tie:
Regarding `document.Iterator`, I can think of two options.

1. We can add a layer of indirection in `database.Table` and keep `document.Iterator` unaware of the context.
   ```
   func (t *Table) Iterator(ctx context.Context) document.IteratorFunc
   func (t *Table) Iterate(ctx context.Context, fn func(d document.Document) error) document.IteratorFunc
   ```

   I’ve tried implementing this approach and it’s pretty easy to lose track of the context flow in `sql/planner`.

2. Since `db.Query` actually returns `query.Result`, we can use `WithContext` pattern (see the previous message). This would also make the API consistent in a sense that only lower-level API accepts explicit context.

asdine:
@tie I think adding context to Genji requires some discussion because it has a very big impact on the project.
I have written an [RFC ](https://github.com/genjidb/genji/pull/249) to express my vision of at least the first iteration of this feature. I think it will be a good place to discuss this and iterate until we find a good solution.

tie:
@asdine I’ll update this PR to conform with the RFC001 and mark it as ready then. Let me know if you want this in a separate PR though.

asdine:
@tie Let's start over on a separate PR if you don't mind

tie:
Sure!

PR #236, opened 2020-10-07
Fixes .Indexes and .tables commands error handler and tests

### LINKED ISSUE #233: Regression in .indexes command
Running `.indexes t` should print `table not found` if table `t` does not exist.

```
--- FAIL: TestRunTablesCmd (0.00s)
    --- FAIL: TestRunTablesCmd/Table_with_options (0.00s)
        command_test.go:38: 
            	Error Trace:	command_test.go:38
            	Error:      	An error is expected but got nil.
            	Test:       	TestRunTablesCmd/Table_with_options
            	Messages:   	%!(EXTRA bool=true)
```

See also: #232

tzzed:
Fixes #233 

`command_test.go` had some errors.
`.Indexes` returns `table not found`

### DIFF
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -52,15 +52,14 @@ func runTablesCmd(db *genji.DB, cmd []string) error {
 
 // displayTableIndex prints all indexes that the given table contains.
 func displayTableIndex(db *genji.DB, tableName string) error {
-	ctx := context.Background()
 	return db.View(func(tx *genji.Tx) error {
+		ctx := context.Background()
 		_, err := tx.QueryDocument(ctx, "SELECT table_name FROM __genji_tables WHERE table_name = ?", tableName)
-		switch err {
-		case database.ErrDocumentNotFound:
-			return database.ErrTableNotFound
-		case nil:
-			// nop
-		default:
+		if err != nil {
+			if err == database.ErrDocumentNotFound {
+				return database.ErrTableNotFound
+			}
+
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


cmd/genji/shell/command.go
```diff
@@ -60,6 +60,11 @@ func displayTableIndex(db *genji.DB, tableName string) error {
 	}
 	defer res.Close()
 
+	_, err = res.Tx.GetTable(tableName)
```
tie, line 63: I think we can select from `__genji_tables`.

https://github.com/genjidb/genji/blob/6aa01a11de96f2da7336b25d38c42d6509e9db69/cmd/genji/shell/shell.go#L435

tzzed: You're it's a better way.

tzzed: You're right it's a better way.

cmd/genji/shell/command.go
```diff
@@ -53,6 +53,15 @@ func runTablesCmd(db *genji.DB, cmd []string) error {
...
+	switch err {
+	case database.ErrDocumentNotFound:
+		return database.ErrTableNotFound
+	case nil:
+		// nop
+	default:
+		return err
+	}
 
 	res, err := db.Query(ctx, "SELECT * FROM __genji_indexes WHERE table_name = ?", tableName)
```
tie, line 66: Both queries should be in the same transaction (using `db.View`).

tzzed: thank you @tie

tie (approved):
LGTM 👍

yaziine (approved):
Thanks for handling this @tzzed, I left just one small comment.

asdine (approved):
Thanks!

Issue #180, opened 2020-09-07
Add support for exit and help commands without leading dot


asdine:
`.help` and `.exit` commands already exist but it's not easy to remember to always use the `.` before. It's especially true for newcomers who want to display the help but don't know what's the syntax.
These commands are among the most used, they should be used intuitively without friction.

```
genji> help
.exit or exit           Exit this program.
.help or help           List all commands.
.indexes [table_name]   Display all indexes or the indexes of the given table name.
.tables                 List names of tables.
genji> exit
<exits the program>
```

PR #239, opened 2020-10-08
#180: add exit and help cmds

ab22:
## Description

Changes for issue #180 

The `help` and `exit` commands work correctly when typing them, however, when typing the new commands the suggestions list won't show up and this is because of the way their [previous commands were added](https://github.com/ab22/genji/blob/master/cmd/genji/shell/command.go#L21-L22).

Also, when typing `.help` or `.exit` it now displays `.help or help` or `.exit or exit` on the suggestions. I am wondering how would you like these commands to be added to the suggestion list. Maybe an `alias []string` under the `commands` structure?

### DIFF
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -18,8 +18,8 @@ var commands = []struct {
 	Options     string
 	Description string
 }{
-	{".exit", ``, "Exit this program."},
-	{".help", ``, "List all commands."},
+	{".exit or exit", ``, "Exit this program."},
+	{".help or help", ``, "List all commands."},
 	{".tables", ``, "List names of tables."},
 	{".indexes", `[table_name]`, "Display all indexes or the indexes of the given table name."},
 	{".dump", `[table_name]`, "Dump database content or table content as SQL statements"},
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -16,13 +16,39 @@ import (
 var commands = []struct {
 	Name        string
 	Options     string
+	DisplayName string
 	Description string
+	Aliases     []string
 }{
-	{".exit or exit", ``, "Exit this program."},
-	{".help or help", ``, "List all commands."},
-	{".tables", ``, "List names of tables."},
-	{".indexes", `[table_name]`, "Display all indexes or the indexes of the given table name."},
-	{".dump", `[table_name]`, "Dump database content or table content as SQL statements"},
+	{
+		Name:        ".exit",
+		DisplayName: ".exit or exit",
+		Description: "Exit this program.",
+		Aliases:     []string{"exit"},
+	},
+	{
+		Name:        ".help",
+		DisplayName: ".help or help",
+		Description: "List all commands.",
+		Aliases:     []string{"help"},
+	},
+	{
+		Name:        ".tables",
+		DisplayName: ".tables",
+		Description: "List names of tables.",
+	},
+	{
+		Name:        ".indexes",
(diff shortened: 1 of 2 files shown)


ab22:
Would you guys like me to propose code changes to correctly support the new command suggestions correctly? @asdine

asdine:
@ab22 Thanks for taking care of this!

> Maybe an alias []string under the commands structure?

I think this is a good idea

cmd/genji/shell/command.go
```diff
@@ -116,26 +134,36 @@ func runHelpCmd() error {
...
+		for _, alias := range c.Aliases {
+			if shouldDisplaySuggestion(alias, in) {
+				suggestions = append(suggestions, alias)
+			}
+		}
 	}
 
 	if len(suggestions) == 0 {
-		return fmt.Errorf("Unknown command %q. Enter \".help\" for help.", in)
+		return fmt.Errorf("unknown command %q. Enter \".help\" for help", in)
```
ab22, line 166: This fixes a go-lint issue: `error strings should not be capitalized or end with punctuation or a newline`

ab22:
I went ahead and added a small command builder. This way we add a little more flexibility for the commands in the future, for example if we wanted to add individual options instead of just a string, then this would easily perform validations when building it.

cmd/genji/shell/command.go
```diff
@@ -13,16 +13,34 @@ import (
...
-	{".exit or exit", ``, "Exit this program."},
-	{".help or help", ``, "List all commands."},
-	{".tables", ``, "List names of tables."},
-	{".indexes", `[table_name]`, "Display all indexes or the indexes of the given table name."},
-	{".dump", `[table_name]`, "Dump database content or table content as SQL statements"},
+	Aliases     []string
+}
+
+var commands = []*command{
+	newCommandBuilder(".exit").
```
asdine, line 25: This feels very Java-esque, I feel the same can be achieved by filling the struct directly:

```go
var commands = []*command{
  {
    name: ".exit",
    aliases: []string{"exit"},
    description: "Exit this program.",
  },
  {
    name: ".help",
    aliases: []string{"help"},
    description: "List all commands.",
  },
}

type command struct {
	Name        string
	Options     string
	DisplayName string
	Description string
}

func (c *command) String() string {
  ...
}
```

cmd/genji/shell/command.go
```diff
@@ -16,13 +16,45 @@ import (
...
+	{
+		Name:        ".help",
+		Options:     "",
+		DisplayName: ".help or help",
+		Description: "List all commands.",
+		Aliases:     []string{"help"},
+	},
+	{
+		Name:        ".tables",
+		Options:     "",
```
asdine, line 39: nit: You don't have to do this in Go, non initialized fields are equal to the [zero value of their type](https://www.geeksforgeeks.org/zero-value-in-golang/)

```suggestion
```

cmd/genji/shell/command.go
```diff
@@ -16,13 +16,45 @@ import (
...
+		DisplayName: ".tables",
+		Description: "List names of tables.",
+		Aliases:     []string{},
```
asdine, line 42: Same here, you can remove this line

asdine (changes requested):
Thanks @ab22, this looks good, I just have a few nitpicking comments

ab22:
Hey @asdine, sure, no problem. I left those out because of how they were previously declared. Just forced pushed a commit to remove those lines. :D

asdine (approved):
Thank you @ab22 !

PR #241, opened 2020-10-08
Add LIKE operator support

tdakkota:
Adds LIKE operator support.
Implementation is based on CockroachDB implementation. 

Updates #52.


sql/query/expr/like.go
```diff
@@ -0,0 +1,193 @@
+package expr
+
+import (
+	"errors"
+	"fmt"
+	"github.com/genjidb/genji/document"
+	"github.com/genjidb/genji/sql/scanner"
+	"regexp"
+	"strings"
+)
```
tie, line 10: We should keep conventional import grouping (`x/tools/cmd/goimports` or `mvdan.cc/gofumpt` may help with this).

```suggestion
import (
	"errors"
	"fmt"
	"regexp"
	"strings"
	
	"github.com/genjidb/genji/document"
	"github.com/genjidb/genji/sql/scanner"
)
```

database/database.go
```diff
@@ -29,6 +30,9 @@ type Database struct {
 	attachedTransaction *Transaction
 	attachedTxMu        sync.Mutex
 
+	// Stores cache of compiled regexp.
+	regexCache regexcache.Cache
```
tie, line 34: I feel like this is over-engineering the problem. We can simply compile the regex when parsing the query, and if it turns out to be terribly slow, add a cache for the whole queries.

/cc @asdine

database/internal/regexcache/list.go
```diff
@@ -0,0 +1,90 @@
+package regexcache
+
+type regexElement struct {
+	next, prev *regexElement
```
tie, line 4: FWIW this looks like doubly linked list to me, which is already available in Go standard library (see [container/list](https://golang.org/pkg/container/list)).

sql/query/expr/comparison.go
```diff
@@ -426,7 +426,7 @@ func (op cmpOp) compare(l, r document.Value) (bool, error) {
 func IsComparisonOperator(op Operator) bool {
 	switch op.(type) {
 	case eqOp, neqOp, gtOp, gteOp, ltOp, lteOp,
-		isOp, isNotOp, inOp, notInOp:
+		isOp, isNotOp, inOp, notInOp, likeOp:
```
tie, line 429: I think it makes sense to add `notLikeOp` here.

database/internal/regexcache/list.go
tdakkota: Type-safety is single reason. 
In general, we can just use https://github.com/hashicorp/golang-lru or any another cache library.

database/database.go
tdakkota: As I know golang regexps can be quite slow.
I think, re-use of compiled regular expressions can be so useful in the common case.
For now, you can use any type of expressions to set the pattern. For example, you can use functions to build the pattern. But if you compile regex when parsing the query, it will be possible to use text literals only.

Also, most simple queries (e.g `LIKE '%abc%'`) do not use cache and regular expressions at all. So, only _complex_ queries will be cached.

I think we can make cache configurable, just allow to disable it.

tie: If we are fine with the existing LIKE syntax, we may write a simple glob matcher. That should take roughly the same amount of code as adding type-safe cache, though a bit more time.

tie: That said, I don’t think regexp compile would cause a noticable performance hit. You may want to write a benchmarks
 (see [Benchmarks in testing package](https://golang.org/pkg/testing/#hdr-Benchmarks) and compare the results with [benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat).

tie: Here is the Go port of the sqlite `icuLikeCompare` routine; it’s about 100 LoC not counting comments.
See also https://goplay.space/#lZPj1yn-O-C

<details>
<summary>like.go</summary>

```
// The author disclaims copyright to this source code.  In place of
// a legal notice, here is a blessing:
//
//    May you do good and not evil.
//    May you find forgiveness for yourself and forgive others.
//    May you share freely, never taking more than you give.
//
// https://www.sqlite.org/src/file?name=ext%2Ficu%2Ficu.c&ln=117-195&m=54b54f02c66c5aea

package expr

import (
	"unicode"
	"unicode/utf8"
)

const (
	matchOne = '_'
	matchAll = '%'
	matchEsc = '\\'
)

func readRune(s string) (rune, string) {
	r, size := utf8.DecodeRuneInString(s)
	if r == utf8.RuneError && size == 1 {
		return rune(s[0]), s[1:]
	}
	return r, s[size:]
}

func skipRune(s string) string {
	_, size := utf8.DecodeRuneInString(s)
	return s[size:]
}

// equalFold is strings.EqualFold for individual runes.
func equalFold(sr, tr rune) bool {
	// Easy case.
	if tr == sr {
		return true
	}

	// Make sr < tr to simplify what follows.
	if tr < sr {
		tr, sr = sr, tr
	}
	// Fast check for ASCII.
	if tr < utf8.RuneSelf {
		// ASCII only, sr/tr must be upper/lower case
		return 'A' <= sr && sr <= 'Z' && tr == sr+'a'-'A'
	}

	// General case. SimpleFold(x) returns the next equivalent rune > x
	// or wraps around to smaller values.
	r := unicode.SimpleFold(sr)
	for r != sr && r < tr {
		r = unicode.SimpleFold(r)
	}
	return r == tr
}

func MatchLike(pattern, s string) bool {
	var prevEscape bool

	for len(pattern) != 0 {
		// Read (and consume) the next character from the input pattern.
		var p rune
		p, pattern = readRune(pattern)

		// There are now 4 possibilities:
		//
		// 1. p is an unescaped match-all character "%",
		// 2. p is an unescaped match-one character "_",
		// 3. p is an unescaped escape character, or
		// 4. p is to be handled as an ordinary character
		//
		if p == matchAll && !prevEscape && p != matchEsc {
			// Case 1.
			var c byte

			// Skip any matchAll or matchOne characters that follow a
			// matchAll. For each matchOne, skip one character in the
			// test string.
			//
			for len(pattern) != 0 {
				c = pattern[0]
				if c != matchAll && c != matchOne {
					break
				}
				pattern = pattern[1:]

				if c != matchOne {
					continue
				}
				if len(s) == 0 {
					return false
				}
				s = skipRune(s)
			}

			if len(pattern) == 0 {
				return true
			}

			for len(s) != 0 {
				if MatchLike(pattern, s) {
					return true
				}
				s = skipRune(s)
			}
			return false
		} else if p == matchOne && !prevEscape && p != matchEsc {
			// Case 2.
			if len(s) == 0 {
				return false
			}
			s = skipRune(s)
		} else if p == matchEsc && !prevEscape {
			// Case 3.
			prevEscape = true
		} else {
			// Case 4.
			var r rune
			r, s = readRune(s)
			if !equalFold(r, p) {
				return false
			}
			prevEscape = false
		}
	}

	return len(s) == 0
}
```
</details>

<details>
<summary>like_test.go</summary>

```
package expr

import (
	"testing"
)

func TestMatchLike(t *testing.T) {
	tests := []struct {
		s, pattern string
		want       bool
	}{
		// Empty
		{"", "", true},
		{"abc", "", false},

		// One
		{"x", "_", true},
		{"xx", "_", false},
		{"", "_", false},

		// Any
		{"abc", "%", true},
		{"", "%", true},

		// Exact
		{"abc", "abc", true},
		{"aBc", "AbC", true},
		{"abc", "def", false},

		// Prefix
		{"abcdef", "abc%", true},
		{"abcdef", "def%", false},

		// Suffix
		{"defabc", "%abc", true},
		{"defabc", "%def", false},

		// Contains
		{"defabcdef", "%abc%", true},
		{"abcd", "%def%", false},

		// Complex
		{"ABCD", "%B%C%", true},
		{"ABCD", "_%B%C%_", true},
		{"ABxCxxD", "a%b%c%d", true},
		{"ABxCxxD", "%__B", false},
	}

	for _, test := range tests {
		if got := MatchLike(test.pattern, test.s); got != test.want {
			t.Errorf(
				"MatchLike(%#v, %#v): expected %#v, got %#v",
				test.pattern, test.s, test.want, got,
			)
		}
	}
}
```
</details>

We can also use their test files.
- https://www.sqlite.org/src/file?name=test/like.test
- https://www.sqlite.org/src/file?name=test/like2.test
- https://www.sqlite.org/src/file?name=test/like3.test

sql/query/expr/like_test.go
```diff
@@ -0,0 +1,51 @@
...
+		want    bool
+		wantErr bool
+	}{
+		{"Empty", args{"", ""}, true, false},
+		{"Empty false", args{"abc", ""}, false, false},
+
+		{"Any", args{"abc", "%"}, true, false},
+		{"Any false", args{"", "%"}, true, false},
+
+		{"Any one and more", args{"abc", "_"}, true, false},
```
tie, line 22: While porting `icuLikeCompare` I also found that this should be `false` (at least in SQLite, not sure about CockroachDB)
See also https://github.com/genjidb/genji/pull/241#discussion_r502796458

database/database.go
tie: FWIW here is the benchmark.

The files `c.txt`, `b.txt` and `a.txt` correspond to `regexp.MatchString`, `(*Regexp).MatchString` and `MatchLike` respectively.

```
go test -count 4 -benchmem -bench Match >a.txt
go test -count 4 -benchmem -bench Regexp >b.txt
go test -count 4 -benchmem -bench Compile >b.txt
sed -i '' -e s:Regexp:Match: -- b.txt
sed -i '' -e s:Compile:Match: -- c.txt
benchstat c.txt b.txt a.txt
```

```
name \ time/op    c.txt        b.txt        a.txt
MatchLike/2-4     4.71µs ± 3%  0.33µs ± 2%  0.16µs ± 0%
MatchLike/3-4     5.63µs ± 4%  0.60µs ± 1%  0.25µs ± 1%
MatchLike/4-4     6.73µs ± 0%  1.01µs ± 1%  0.33µs ± 1%
MatchLike/5-4     8.44µs ± 2%  1.41µs ± 2%  0.41µs ± 1%
MatchLike/6-4     9.73µs ± 2%  1.96µs ± 2%  0.49µs ± 0%
MatchLike/7-4     11.0µs ± 1%   2.5µs ± 0%   0.6µs ± 0%
MatchLike/8-4     12.6µs ± 1%   3.3µs ± 3%   0.7µs ± 1%

name \ alloc/op   c.txt        b.txt        a.txt
MatchLike/2-4     4.04kB ± 0%  0.00kB       0.00kB     
MatchLike/3-4     4.77kB ± 0%  0.00kB       0.00kB     
MatchLike/4-4     5.56kB ± 0%  0.00kB       0.00kB     
MatchLike/5-4     7.62kB ± 0%  0.00kB       0.00kB     
MatchLike/6-4     8.24kB ± 0%  0.00kB       0.00kB     
MatchLike/7-4     9.09kB ± 0%  0.00kB       0.00kB     
MatchLike/8-4     10.0kB ± 0%   0.0kB        0.0kB     

name \ allocs/op  c.txt        b.txt        a.txt
MatchLike/2-4       41.0 ± 0%     0.0          0.0     
MatchLike/3-4       44.0 ± 0%     0.0          0.0     
MatchLike/4-4       49.0 ± 0%     0.0          0.0     
MatchLike/5-4       53.0 ± 0%     0.0          0.0     
MatchLike/6-4       56.0 ± 0%     0.0          0.0     
MatchLike/7-4       59.0 ± 0%     0.0          0.0     
MatchLike/8-4       64.0 ± 0%     0.0          0.0     
```

<details>
<summary>benchmark_test.go</summary>

```
package glob

import (
	"regexp"
	"strconv"
	"strings"
	"testing"
)

func BenchmarkMatchLike(b *testing.B) {
	benchmarkLike(b, func(p string) func(string) bool {
		return func(s string) bool {
			return MatchLike(p, s)
		}
	})
}

func BenchmarkRegexpLike(b *testing.B) {
	benchmarkLike(b, func(p string) func(string) bool {
		p = strings.Replace(p, "%", ".*", -1)
		p = strings.Replace(p, "_", ".", -1)
		p = "^" + p + "$"
		re := regexp.MustCompile(p)
		return re.MatchString
	})
}

func BenchmarkCompileLike(b *testing.B) {
	benchmarkLike(b, func(p string) func(string) bool {
		p = strings.Replace(p, "%", ".*", -1)
		p = strings.Replace(p, "_", ".", -1)
		p = "^" + p + "$"
		return func(s string) bool {
			re := regexp.MustCompile(p)
			return re.MatchString(s)
		}
	})
}

func benchmarkLike(b *testing.B, fn func(p string) func(string) bool) {
	for n := 2; n <= 8; n++ {
		p := strings.Repeat("%z", n)
		s := strings.Repeat("xyz", n)
		m := fn(p)

		b.Run(strconv.Itoa(n), func(b *testing.B) {
			for i := 0; i < b.N; i++ {
				m(s)
			}
		})
	}
}
```
</details>

tie: One more thing to note is that we may want to limit the length of the pattern—SQLite limits pattern length to 2¹⁶/50000 bytes, which is more than enough.

sql/query/glob/like.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+			s = skipRune(s)
+		} else if p == matchEsc && !prevEscape {
+			// Case 3.
+			prevEscape = true
+		} else {
+			// Case 4.
+			var r rune
+			r, s = readRune(s)
+			if !equalFold(p, r) {
```
tdakkota, line 132: Shouldn't we use direct comparison?

I ran this query on different DBs.
```SQL
SELECT 'abc' LIKE 'ABC';
```
Results:
- PostgreSQL 12.3 returns false.
- SQLite 3.27.2 returns 1 (true).
- MySQL 5.7.12 returns 1 (true).
- Oracle Database 11g returns false (Query: `SELECT * FROM DUAL WHERE 'abc' like 'ABC'`, should be non-empty if true).

database/database.go
tie: OK, we can do even better than SQLite and use backtracking. This also removes stack overflow issue for large patterns, and more test cases.

<details>
<summary>like.go</summary>

```
// The author disclaims copyright to this source code.  In place of
// a legal notice, here is a blessing:
//
//    May you do good and not evil.
//    May you find forgiveness for yourself and forgive others.
//    May you share freely, never taking more than you give.
//
// https://sqlite.org/src/file?name=ext%2Ficu%2Ficu.c&ln=117-195&m=54b54f02c66c5aea

package glob

import (
	"unicode"
	"unicode/utf8"
)

const (
	matchOne = '_'
	matchAll = '%'
	matchEsc = '\\'
)

// readRune is like skipRune, but also returns the removed Unicode code point.
func readRune(s string) (rune, string) {
	if s == "" {
		panic("")
	}
	r, size := utf8.DecodeRuneInString(s)
	if r == utf8.RuneError && size == 1 {
		return rune(s[0]), s[1:]
	}
	return r, s[size:]
}

// skipRune returns a slice of the string s with the first Unicode code point removed.
func skipRune(s string) string {
	if s == "" {
		panic("")
	}
	_, size := utf8.DecodeRuneInString(s)
	return s[size:]
}

// equalFold is strings.EqualFold for individual runes.
func equalFold(sr, tr rune) bool {
	// Easy case.
	if tr == sr {
		return true
	}

	// Make sr < tr to simplify what follows.
	if tr < sr {
		tr, sr = sr, tr
	}
	// Fast check for ASCII.
	if tr < utf8.RuneSelf {
		// ASCII only, sr/tr must be upper/lower case
		return 'A' <= sr && sr <= 'Z' && tr == sr+'a'-'A'
	}

	// General case. SimpleFold(x) returns the next equivalent rune > x
	// or wraps around to smaller values.
	r := unicode.SimpleFold(sr)
	for r != sr && r < tr {
		r = unicode.SimpleFold(r)
	}
	return r == tr
}

// MatchLike reports whether string s matches the SQL LIKE-style glob pattern.
// Supported wildcards are '_' (match any one character) and '%' (match zero
// or more characters). They can be escaped by '\' (escape character).
//
// MatchLike requires pattern to match whole string, not just a substring.
func MatchLike(pattern, s string) bool {
	var prevEscape bool

	var w, t string // backtracking state

	for len(s) != 0 {
		if len(pattern) == 0 {
			return false
		}

		// Read (and consume) the next character from the input pattern.
		var p rune
		p, pattern = readRune(pattern)

	backtrack:
		// There are now 4 possibilities:
		//
		// 1. p is an unescaped match-all character "%",
		// 2. p is an unescaped match-one character "_",
		// 3. p is an unescaped escape character, or
		// 4. p is to be handled as an ordinary character
		//
		if p == matchAll && !prevEscape && p != matchEsc {
			// Case 1.
			var c byte

			// Skip any matchAll or matchOne characters that follow a
			// matchAll. For each matchOne, skip one character in the
			// test string.
			//
			for len(pattern) != 0 {
				c = pattern[0]
				if c != matchAll && c != matchOne {
					break
				}
				pattern = pattern[1:]

				if c != matchOne {
					continue
				}
				if len(s) == 0 {
					return false
				}
				s = skipRune(s)
			}

			if len(pattern) == 0 {
				return true
			}

			// Save state and match next character.
			//
			w, t = pattern, s
		} else if p == matchOne && !prevEscape && p != matchEsc {
			// Case 2.
			if len(s) == 0 {
				return false
			}
			s = skipRune(s)
		} else if p == matchEsc && !prevEscape {
			// Case 3.
			prevEscape = true
		} else {
			// Case 4.
			prevEscape = false

			var r rune
			r, s = readRune(s)
			if !equalFold(p, r) {
				if len(w) == 0 {
					// Nothing to backtrack.
					return false
				}
				// Keep the pattern and skip rune in input.
				// Note that we only backtrack to matchAll.
				p, pattern = matchAll, w
				s = skipRune(t)
				goto backtrack
			}
		}
	}

	// Check that the rest of the pattern is matchAll.
	for i := 0; i < len(pattern); i++ {
		if pattern[i] == matchAll {
			continue
		}

		// Allow escaping end of string.
		if i+1 == len(pattern) {
			if pattern[i] == matchEsc {
				return true
			}
		}

		return false
	}
	return true
}
```
</details>

<details>
<summary>like_test.go</summary>

```
package glob

import (
	"testing"
)

func TestMatchLike(t *testing.T) {
	tests := []struct {
		s, pattern string
		want       bool
	}{
		// Empty
		{"", "", true},
		{"", "x", false},
		{"x", "", false},

		// One
		{"", "_", false},
		{"x", "_", true},
		{"x", "__", false},
		{"xx", "_", false},
		{"bLah", "bL_h", true},
		{"bLaaa", "bLa_", false},
		{"bLah", "bLa_", true},
		{"bLaH", "_Lah", true},
		{"bLaH", "_LaH", true},

		// All
		{"", "%", true},
		{"abc", "%", true},
		{"abc", "%%", true},

		// Any and one
		{"x", "%_", true},
		{"", "%_", false},

		// Escape
		{"", "\\", true},
		{"%", "\\%", true},
		{"_", "\\_", true},
		{"x", "\\%", false},
		{"x", "\\_", false},
		{"x", "\\x", true},

		// Escaping escape
		{"\\", "\\\\", true},
		{"\\", "\\\\%", true},
		{"\\", "\\\\_", false},
		{"\\x", "\\\\x", true},

		// Exact
		{"abc", "abc", true},
		{"aBc", "AbC", true},
		{"abc", "def", false},

		// Case folding
		{"K", "\u212A", true}, // K → k → U+212A
		{"\u212A", "k", true},

		// Invalid UTF-8
		{"\xFF", "\xFF", true},
		{"\xFA", "\xFB", false},
		{"\xFF", "_", true},
		{"\xFF", "\xFF_", false},
		{"\xFF", "%", true},
		{"\xFF", "%\xFF%", true},
		{"\xFF", "x", false},

		// Prefix
		{"abc", "abc%", true},
		{"abcdef", "abc%", true},
		{"abcdef", "def%", false},

		// Suffix
		{"abc", "%abc", true},
		{"defabc", "%abc", true},
		{"defabc", "%def", false},

		// Contains
		{"defabcdef", "%abc%", true},
		{"abcd", "%def%", false},
		{"abc", "b", false},

		// Complex
		{"abc", "ab%d", false},
		{"ABCD", "%B%C%", true},
		{"ABxCxxD", "a%b%c%d", true},
		{"a", "__", false},
		{"ab", "__", true},
		{"abc", "___", true},
		{"abcd", "____", true},
		{"abc", "____", false},
		{"abcd", "_b__", true},
		{"abcd", "_a__", false},
		{"abcd", "__c_", true},
		{"abcd", "__d_", false},

		// Mixed
		{"", "%_", false},
		{"", "_%", false},
		{"a", "%_", true},
		{"a", "%__", false},
		{"ab", "%_", true},
		{"abc", "%_", true},
		{"ab", "_%_", true},
		{"ab", "%_%_%", true},
		{"aaaa", "_aa%", true},
		{"abc", "_%%_%_", true},
		{"abc", "_%%_%&_", false},
		{"abcd", "_b%__", true},
		{"abcd", "_a%__", false},
		{"abcd", "_%%_c_", true},
		{"abcd", "_%%_d_", false},
		{"abcde", "_b_d%_", true},
		{"abcde", "_%b%_%d%_", true},
		{"abcd", "_%b%c%_", true},
		{"ABxCxxD", "%__B", false},
		{"abBbc", "%b_c", true},

		// Longer strings
		{
			"%abc%",
			"%%\\%a%b%c\\%%%",
			true,
		},
		{
			"aaabbaabbaab",
			"%aabbaa%a%",
			true,
		},
		{
			"abacaaadabacababacaaabadagabacaba",
			"%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%",
			true,
		},
		{
			"aaaaaaaaaaaaaaaa",
			"%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%",
			false,
		},
		{
			"%a%b%c%",
			"%%%%%%%%a%%%%\\%%%%b%%%%\\%%%%c%%%%%%%%",
			true,
		},
		{
			"a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%a%",
			"a%a\\%a%a\\%a%a\\%a%a\\%a%a\\%a%a\\%a%a\\%a%a\\%a%",
			true,
		},
		{
			"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaab",
			"a%a%a%a%a%a%aa%aaa%a%a%b",
			true,
		},
		{
			"abababababababababababababababababababaacacacacacacacadaeafagahaiajakalaaaaaaaaaaaaaaaaaffafagaagggagaaaaaaaab",
			"%a%b%ba%ca%a%aa%aaa%fa%ga%b%",
			true,
		},
		{
			"abababababababababababababababababababaacacacacacacacadaeafagahaiajakalaaaaaaaaaaaaaaaaaffafagaagggagaaaaaaaab",
			"%a%b%ba%ca%a%x%aaa%fa%ga%b%",
			false,
		},
		{
			"abababababababababababababababababababaacacacacacacacadaeafagahaiajakalaaaaaaaaaaaaaaaaaffafagaagggagaaaaaaaab",
			"%a%b%ba%ca%aaaa%fa%ga%gggg%b%",
			false,
		},
		{
			"abababababababababababababababababababaacacacacacacacadaeafagahaiajakalaaaaaaaaaaaaaaaaaffafagaagggagaaaaaaaab",
			"%a%b%ba%ca%aaaa%fa%ga%ggg%b%",
			true,
		},
	}

	for _, test := range tests {
		if got := MatchLike(test.pattern, test.s); got != test.want {
			t.Errorf(
				"MatchLike(%#v, %#v): expected %#v, got %#v",
				test.pattern, test.s, test.want, got,
			)
		}
	}
}
```
</details>

```
name            old time/op    new time/op    delta
MatchLike/0-4     85.5ns ± 1%    91.5ns ± 9%   +7.03%  (p=0.000 n=7+8)
MatchLike/1-4      165ns ± 0%     170ns ± 1%   +3.33%  (p=0.000 n=7+8)
MatchLike/2-4      337ns ± 1%     351ns ± 2%   +4.19%  (p=0.001 n=7+7)
MatchLike/3-4      666ns ± 0%     674ns ± 1%   +1.23%  (p=0.000 n=7+8)
MatchLike/4-4     1.34µs ± 1%    1.34µs ± 1%     ~     (p=0.859 n=8+8)
MatchLike/5-4     2.65µs ± 1%    2.66µs ± 3%     ~     (p=0.588 n=8+7)
MatchLike/6-4     5.26µs ± 1%    5.31µs ± 1%   +1.02%  (p=0.007 n=8+8)
MatchLike/7-4     10.6µs ± 1%    10.7µs ± 2%     ~     (p=0.487 n=8+8)
MatchLike/8-4     21.3µs ± 1%    21.3µs ± 1%     ~     (p=0.536 n=8+7)
MatchLike/9-4     43.0µs ± 1%    42.3µs ± 0%   -1.74%  (p=0.001 n=7+7)
MatchLike/10-4    85.8µs ± 1%    84.4µs ± 1%   -1.65%  (p=0.001 n=7+8)
MatchLike/11-4     174µs ± 2%     169µs ± 0%   -3.43%  (p=0.000 n=8+8)
MatchLike/12-4     360µs ± 1%     338µs ± 1%   -6.05%  (p=0.000 n=8+8)
MatchLike/13-4     742µs ± 3%     680µs ± 2%   -8.38%  (p=0.000 n=8+8)
MatchLike/14-4    1.47ms ± 2%    1.35ms ± 0%   -8.18%  (p=0.000 n=8+7)
MatchLike/15-4    3.28ms ±10%    2.70ms ± 0%  -17.80%  (p=0.000 n=8+8)
```

sql/query/glob/like.go
tie: I think we should stick with what MySQL and SQLite do.

sql/query/glob/like.go
```diff
@@ -0,0 +1,169 @@
...
+	backtrack:
+		if len(w) == 0 {
+			// Nothing to backtrack.
+			return prevEscape
+		}
+		// Keep the pattern and skip rune in input.
+		// Note that we only backtrack to matchAll.
+		p, pattern = matchAll, w
+		s = skipRune(t)
+		goto loop
```
tie, line 150: Oops, we have to reset `prevEscape` here.

```suggestion
		p, pattern = matchAll, w
		s = skipRune(t)
		prevEscape = false
		goto loop
```

sql/query/glob/like.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+
+			// Save state and match next character.
+			//
+			w, t = pattern, s
+		} else if p == matchOne && !prevEscape {
+			// Case 2.
+			if len(s) == 0 {
+				return false
+			}
```
tie, line 124: Looks like this block is unreachable.

```suggestion
			//
			// We can either reach this case on normal iteration where len(s) ≠ 0,
			// or from backtracking. But we consume all match-one characters before
			// saving backtracking state, so it’s only reachable on normal iteration.
			// That is, we are guaranteed to have input at this point.
			//
```

sql/query/glob/like.go
```diff
@@ -0,0 +1,169 @@
...
+		if len(pattern) == 0 {
+			goto backtrack
+		}
+		p, pattern = readRune(pattern)
+
+		// There are now 4 possibilities:
+		//
+		// 1. p is an unescaped match-all character "%",
+		// 2. p is an unescaped match-one character "_",
+		// 3. p is an unescaped escape character, or
```
tie, line 86:
```suggestion
		// 1. p is an unescaped matchAll character “%”,
		// 2. p is an unescaped matchOne character “_”,
		// 3. p is an unescaped matchEsc character, or
```

sql/query/glob/like.go
tie: In fact, ideally we should be comparing grapheme clusters using e.g. [github.com/clipperhouse/uax29/graphemes](https://godoc.org/github.com/clipperhouse/uax29/graphemes).

sql/query/glob/like.go
tie: @tdakkota I’ve created https://github.com/tdakkota/genji/pull/2 with these changes.

asdine (commented):
Thanks! That's some awesome work 👏🏼 
The PR is still in draft, what do you think is missing to finalize it?

sql/query/glob/like.go
tie: SQLite performs (simple) case folding for character comparison with `LIKE` operator, and since that’s what our implementation is based on, I think it’s reasonable to follow this behavior too.

A proper Unicode support would definitely take some time to implement given the current state of Unicode support in Go (scattered across third-party libraries with different Unicode versions, and each embeds their own character database copy).

tie:
@tdakkota can you please mark outdated discussions as resolved?

asdine:
@tdakkota Could you give us some input on what's missing?

tie:
@asdine I think it’s ready for review, otherwise the PR would be still a draft.

>@tdakkota marked this pull request as ready for review 7 days ago

asdine (approved):
Thanks!

PR #242, opened 2020-10-08
Add ADD COLUMN statement, add DEFAULT field constraint

### LINKED ISSUE #96: Add ALTER TABLE ... ADD FIELD ...
This is a sub-issue of #94 that focuses on adding a new field constraint.

We should be able to do the following:

```
genji> CREATE TABLE foo;
genji> INSERT INTO foo VALUES {name: "John Doe"};
genji> ALTER TABLE foo ADD FIELD age int;
```
If the field `age` already exists, we should return an error.

#### NOT NULL 
The only constraint supported for now would be `NOT NULL` but a default value must be specified for the existing documents.
```
genji> ALTER TABLE foo ADD FIELD age int NOT NULL DEFAULT 50;
genji> SELECT * FROM foo;
{
  "name": "John Doe",
  "age": 50
}
```

#### Dot notation
We should also support the [dot notation](https://genji.dev/docs/genji-sql/lexical-structure/#dot-notation) structure.
For example:
```
genji> ALTER TABLE foo ADD FIELD adress.zip int;
genji> INSERT INTO foo VALUES {name: "John Wick", age: 50, adress: {zip:75000}};
genji> SELECT * FROM foo;
{
  "name": "John Wick",
  "age": 50,
  "adress": {
    "zip": 75000
  }
}
genji> INSERT INTO foo VALUES {name: "John Wick", age: 50, adress: {zip:"Paris"}};
cannot convert "text" to "int64": type "text" incompatible with "integer"
```

The same goes for arrays:
```
genji> ALTER TABLE foo ADD FIELD animals.0 text;
genji> INSERT INTO foo VALUES {name: "John Lennon", age: 50, animals: ["bird"]};
genji> SELECT * FROM foo;
{
  "name": "John Lennon",
  "age": 50,
  "animals": [
    "cat",
  ]
}
genji> INSERT INTO foo VALUES {name: "John Lennon", age: 50, animals: [127]};
cannot convert "int8" to "string"
```

tdakkota:
Fixes #96.

Adds:
- New `DEFAULT` constraint which sets default value of field/column.
- `ALTER TABLE <table_name> ADD COLUMN <field definition>` statement to add new fields to the table. Original issue requests `ADD FIELD` statement, but I think `ADD COLUMN` is more compatible with SQL. Anyway, it's not too hard to fix.

I have some questions about the _correct_ table info modification.

### DIFF
--- a/database/config.go
+++ b/database/config.go
@@ -319,13 +319,7 @@ func (t *tableInfoStore) modifyTable(tx *Transaction, tableName string, f func(*
 		return err
 	}
 
-	key := []byte(tableName)
-	err = st.Delete(key)
-	if err != nil {
-		return err
-	}
-
-	err = st.Put(key, buf.Bytes())
+	err = st.Put([]byte(tableName), buf.Bytes())
 	if err != nil {
 		return err
 	}
--- a/database/config.go
+++ b/database/config.go
@@ -18,6 +18,11 @@ type FieldConstraint struct {
 	Type         document.ValueType
 	IsPrimaryKey bool
 	IsNotNull    bool
+	DefaultValue document.Value
+}
+
+func (f *FieldConstraint) HasDefaultValue() bool {
+	return f.DefaultValue.Type != 0
 }
 
 // ToDocument returns a document from f.
@@ -28,6 +33,7 @@ func (f *FieldConstraint) ToDocument() document.Document {
 	buf.Add("type", document.NewIntegerValue(int64(f.Type)))
 	buf.Add("is_primary_key", document.NewBoolValue(f.IsPrimaryKey))
 	buf.Add("is_not_null", document.NewBoolValue(f.IsNotNull))
+	buf.Add("default_value", f.DefaultValue)
 	return buf
 }
 
@@ -60,6 +66,13 @@ func (f *FieldConstraint) ScanDocument(d document.Document) error {
 		return err
 	}
 	f.IsNotNull = v.V.(bool)
+
+	v, err = d.GetByField("default_value")
+	if err != nil {
+		return err
+	}
+	f.DefaultValue = v
+
 	return nil
 }
 
@@ -319,7 +332,7 @@ func (t *tableInfoStore) loadAllTableInfo(tx engine.Transaction) error {
 
 	t.tableInfos[indexStoreName] = TableInfo{
 		storeName: []byte(indexStoreName),
-		readOnly: true,
+		readOnly:  true,
 		FieldConstraints: []FieldConstraint{
 			{
 				Path: document.ValuePath{
--- a/database/config.go
+++ b/database/config.go
@@ -33,7 +33,9 @@ func (f *FieldConstraint) ToDocument() document.Document {
 	buf.Add("type", document.NewIntegerValue(int64(f.Type)))
 	buf.Add("is_primary_key", document.NewBoolValue(f.IsPrimaryKey))
 	buf.Add("is_not_null", document.NewBoolValue(f.IsNotNull))
-	buf.Add("default_value", f.DefaultValue)
+	if f.HasDefaultValue() {
+		buf.Add("default_value", f.DefaultValue)
+	}
 	return buf
 }
 
@@ -68,10 +70,12 @@ func (f *FieldConstraint) ScanDocument(d document.Document) error {
 	f.IsNotNull = v.V.(bool)
 
 	v, err = d.GetByField("default_value")
-	if err != nil {
+	if err != nil && err != document.ErrFieldNotFound {
 		return err
 	}
-	f.DefaultValue = v
+	if err == nil {
+		f.DefaultValue = v
+	}
 
 	return nil
 }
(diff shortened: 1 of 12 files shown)


sql/parser/alter.go
```diff
@@ -5,40 +5,67 @@ import (
...
+}
+
+func (p *Parser) parseAlterTableAddFieldStatement(tableName string) (_ query.AlterTableAddColumn, err error) {
+	var stmt query.AlterTableAddColumn
+	stmt.TableName = tableName
+
+	// Parse "COLUMN".
+	if tok, pos, lit := p.ScanIgnoreWhitespace(); tok != scanner.COLUMN {
+		return stmt, newParseError(scanner.Tokstr(tok, lit), []string{"COLUMN"}, pos)
+	}
```
tie, line 33: >Original issue requests ADD FIELD statement, but I think ADD COLUMN is more compatible with SQL. Anyway, it's not too hard to fix.

Perhaps we can allow both keywords and use the AddField in internal API (`s/AddColumn/AddField`)?

database/transaction.go
```diff
@@ -128,6 +128,49 @@ func (tx *Transaction) GetTable(name string) (*Table, error) {
 	}, nil
 }
 
+// modifyTable modifies TableInfo using given callback.
+func (tx *Transaction) modifyTable(name string, f func(*TableInfo) error) error {
```
asdine, line 132: This should be a method on `tableInfoStore` because we want to do that update atomically, using the provided mutex.
The problem here is that we can have race conditions between the call to `tableInfoStore.Get`, `tableInfoStore.Delete` and `tableInfoStore.Insert`.

sql/parser/create.go
```diff
@@ -75,6 +76,17 @@ func (p *Parser) parseIfNotExists() (bool, error) {
 	return true, nil
 }
 
+func (p *Parser) parseField(fc *database.FieldConstraint) (err error) {
+	fc.Path, err = p.parsePath()
+	if err != nil {
+		p.Unscan()
```
asdine, line 82: Do we want to unscan here? We usually don't on error

sql/parser/create.go
```diff
@@ -75,6 +76,17 @@ func (p *Parser) parseIfNotExists() (bool, error) {
 	return true, nil
 }
 
+func (p *Parser) parseField(fc *database.FieldConstraint) (err error) {
```
asdine, line 79: How about `parseFieldDefinition` or something like that? I feel like `parseField` is too broad and doesn't really describe what the function is doing.

asdine (commented):
Thanks @tdakkota this is great work! It's great to see that you seem comfortable with the codebase!

database/config.go
```diff
@@ -319,7 +332,7 @@ func (t *tableInfoStore) loadAllTableInfo(tx engine.Transaction) error {
 
 	t.tableInfos[indexStoreName] = TableInfo{
 		storeName: []byte(indexStoreName),
-		readOnly: true,
+		readOnly:  true,
```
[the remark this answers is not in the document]
tie: Looks like a `gofmt` formatting :) I think it’s fine to leave it as is, unless we want this line to keep popping up in each PR that changes this file.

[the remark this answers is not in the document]
asdine: My bad, I thought it was added by mistake

sql/parser/alter_test.go
```diff
@@ -34,3 +36,59 @@ func TestParserAlterTable(t *testing.T) {
...
+				Type: document.IntegerValue,
+			},
+		}, false},
+		{"With not null", "ALTER TABLE foo ADD FIELD bar NOT NULL", query.AlterTableAddField{TableName: "foo",
+			Constraint: database.FieldConstraint{
+				Path:      parsePath(t, "bar"),
+				IsNotNull: true,
+			},
+		}, false},
+		{"With primary key", "ALTER TABLE foo ADD FIELD bar PRIMARY KEY", query.AlterTableAddField{TableName: "foo",
```
yaziine, line 64: A `PRIMARY KEY` field should hold unique value therefore we shouldn't support altering a table by adding a `PRIMARY KEY` constraint. We don't have a way to update all existing documents with different values.

yaziine (changes requested):

yaziine (changes requested):
Thanks @tdakkota for doing this! 🎉 

I only have one concern regarding the `PRIMARY KEY` constraint otherwise everything looks good. 👍

sql/parser/alter_test.go
asdine: Good catch @yaziine! Indeed, we don't have a strategy for changing the primary key yet. We should not allow this for now

sql/parser/alter.go
```diff
@@ -38,6 +38,10 @@ func (p *Parser) parseAlterTableAddFieldStatement(tableName string) (_ query.Alt
 		return stmt, err
 	}
 
+	if stmt.Constraint.IsPrimaryKey {
+		return stmt, &ParseError{Message: "Adding a field with primary key constraint is not supported now."}
```
yaziine, line 42: This is more consistent with other errors in this package and I am not sure that we'll even support this in the future.
```suggestion
		return stmt, &ParseError{Message: "cannot add a PRIMARY KEY constraint"}
```

yaziine (approved):

database/config.go
```diff
@@ -285,6 +289,53 @@ func (t *tableInfoStore) Delete(tx *Transaction, tableName string) error {
...
+	err = t.db.Codec.NewEncoder(&buf).EncodeDocument(info.ToDocument())
+	if err != nil {
+		return err
+	}
+
+	key := []byte(tableName)
+	err = st.Delete(key)
+	if err != nil {
+		return err
+	}
```
asdine, line 326: This is not necessary, as Put is a replace operation

```suggestion
```

asdine (approved):
Thanks! It's ready to merge after conflict handling 👍🏼

asdine:
@tdakkota Could you rebase your branch?

Issue #237, opened 2020-10-07
Display DOUBLES  with a decimal point


tzzed:
Doubles with `.0` decimal is currently printed as `integer` it should be displayed as double.
```SQL
CREATE TABLE foo (a DOUBLE NOT NULL, b DOUBLE NOT NULLl);
INSERT INTO foo VALUES {"a": 2.0, "b": 3.1};
SELECT * from foo;
{
  "a": 2,
  "b": 3.1
}
```

expected 

```SQL 
SELECT * from foo;
{
  "a": 2.0,
  "b": 3.1
}
```

Amirhossein2000:
Hi
I want to work on this issue.
I think you want this only in Genji shell, true?

asdine:
It must be done in the document package. Basically there is a special type named [jsonDocument](https://github.com/genjidb/genji/blob/5a3cc7be3783457ed96050d7e7c1369efbaadafb/document/document.go#L430-L463) that manually turns a document into json.
Then it calls the [MarshalJSON ](https://github.com/genjidb/genji/blob/5a3cc7be3783457ed96050d7e7c1369efbaadafb/document/value.go#L243:16) method of each value, which might be the culprit.

Amirhossein2000:
It's Done now,
I used this https://stackoverflow.com/questions/52446730/stop-json-marshal-from-stripping-trailing-zero-from-floating-point-number/52446854 to fix this.

PR #246, opened 2020-10-09
Display doubles with a decimal point

Amirhossein2000:
This pull request is for #237

### DIFF
--- a/document/value.go
+++ b/document/value.go
@@ -250,15 +250,10 @@ func (v Value) MarshalJSON() ([]byte, error) {
 		return strconv.AppendInt(nil, v.V.(int64), 10), nil
 	case DoubleValue:
 		f := v.V.(float64)
-		abs := math.Abs(f)
-		fmt := byte('f')
-		if abs != 0 {
-			if abs < 1e-6 || abs >= 1e21 {
-				fmt = 'e'
-			}
+		if f == float64(int(f)) {
+			return strconv.AppendFloat(nil, f, 'f', 1, 64), nil
 		}
-
-		return strconv.AppendFloat(nil, v.V.(float64), fmt, -1, 64), nil
+		return strconv.AppendFloat(nil, f, 'f', -1, 64), nil
 	case TextValue:
 		return []byte(strconv.Quote(v.V.(string))), nil
 	case BlobValue:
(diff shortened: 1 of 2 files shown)


document/value.go
```diff
@@ -250,15 +250,10 @@ func (v Value) MarshalJSON() ([]byte, error) {
...
-		if abs != 0 {
-			if abs < 1e-6 || abs >= 1e21 {
-				fmt = 'e'
-			}
+		if float64(f) == float64(int(f)) {
+			return []byte(strconv.FormatFloat(float64(f), 'f', 1, 64)), nil
 		}
-
-		return strconv.AppendFloat(nil, v.V.(float64), fmt, -1, 64), nil
+		return []byte(strconv.FormatFloat(float64(f), 'f', -1, 64)), nil
```
asdine, line 256: This allocates the formatted value twice, because the string returned by `FormatFloat` is then copied when it's casted to a byte slice. We should use `strconv.AppendFloat` as it's only allocating that memory once.
This is important because this will be called thousands of times per second.

asdine (changes requested):
Thanks @Amirhossein2000 ! I have one comment regarding performance

document/value.go
Amirhossein2000: Done

document/value.go
```diff
@@ -250,15 +250,10 @@ func (v Value) MarshalJSON() ([]byte, error) {
 		return strconv.AppendInt(nil, v.V.(int64), 10), nil
 	case DoubleValue:
 		f := v.V.(float64)
-		abs := math.Abs(f)
-		fmt := byte('f')
-		if abs != 0 {
-			if abs < 1e-6 || abs >= 1e21 {
-				fmt = 'e'
-			}
+		if float64(f) == float64(int(f)) {
```
asdine, line 253: f is already a float64, I think we avoid casting it here and below for clarity

Amirhossein2000: Done

document/value.go
```diff
@@ -250,15 +250,10 @@ func (v Value) MarshalJSON() ([]byte, error) {
 		return strconv.AppendInt(nil, v.V.(int64), 10), nil
 	case DoubleValue:
 		f := v.V.(float64)
-		abs := math.Abs(f)
-		fmt := byte('f')
-		if abs != 0 {
-			if abs < 1e-6 || abs >= 1e21 {
-				fmt = 'e'
-			}
+		if f == f {
```
asdine, line 253: It should be `f == float64(int(f))` isn't it?

Amirhossein2000: Done

asdine (approved):
Thank you!

PR #247, opened 2020-10-09
Explicit message error for table that doesn't exist

carlosfrodrigues:
The solution I gave was change the variable ErrTableNotFound to a  function with the same name that receives the table name as a string and concatenate it in the error message.

### DIFF
--- a/cmd/genji/shell/command.go
+++ b/cmd/genji/shell/command.go
@@ -16,13 +16,39 @@ import (
 var commands = []struct {
 	Name        string
 	Options     string
+	DisplayName string
 	Description string
+	Aliases     []string
 }{
-	{".exit or exit", ``, "Exit this program."},
-	{".help or help", ``, "List all commands."},
-	{".tables", ``, "List names of tables."},
-	{".indexes", `[table_name]`, "Display all indexes or the indexes of the given table name."},
-	{".dump", `[table_name]`, "Dump database content or table content as SQL statements"},
+	{
+		Name:        ".exit",
+		DisplayName: ".exit or exit",
+		Description: "Exit this program.",
+		Aliases:     []string{"exit"},
+	},
+	{
+		Name:        ".help",
+		DisplayName: ".help or help",
+		Description: "List all commands.",
+		Aliases:     []string{"help"},
+	},
+	{
+		Name:        ".tables",
+		DisplayName: ".tables",
+		Description: "List names of tables.",
+	},
+	{
+		Name:        ".indexes",
+		Options:     "[table_name]",
+		DisplayName: ".indexes",
+		Description: "Display all indexes or the indexes of the given table name.",
+	},
+	{
+		Name:        ".dump",
+		Options:     "[table_name]",
+		DisplayName: ".dump",
+		Description: "Dump database content or table content as SQL statements.",
+	},
 }
 
 // runTablesCmd shows all tables.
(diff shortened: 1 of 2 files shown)


carlosfrodrigues:
Someone can help me? It's failing only in two tests: 
drop_test.go:27
reindex_test.go:53

database/errors.go
```diff
@@ -2,12 +2,15 @@ package database
 
 import (
 	"errors"
+	"fmt"
 )
 
-var (
-	// ErrTableNotFound is returned when the targeted table doesn't exist.
-	ErrTableNotFound = errors.New("table not found")
```
tie, line 9: I’d suggest to keep the error variable and call `fmt.Errorf("%w: %q", ErrTableNotFound, tableName)` at each fail point.

Then you’d need to replace all `err == ErrTableNotFound ` comparisons with `errors.Is(err, ErrTableNotFound)`.

database/transaction_test.go
```diff
@@ -79,7 +79,7 @@ func TestTxTable(t *testing.T) {
 
 		// Getting a table that doesn't exist should fail.
 		_, err = tx.GetTable("unknown")
-		require.EqualError(t, err, database.ErrTableNotFound.Error())
```
tie, line 82: If we want to check that a particular table does not exist, it’d be better to define the following type

```
type TableNotFoundError struct {
	Name string
}

func (*TableNotFoundError) Is(err error) bool {
	return err == ErrTableNotFound
}
```

Then unwrap the error:

```
var notFoundError TableNotFoundError
if errors.As(err, &notFoundError) {
	fmt.Println(notFoundError.Name)
}
```

database/errors.go
```diff
@@ -2,12 +2,15 @@ package database
...
+// ErrTableNotFound is returned when the targeted table doesn't exist.
+func ErrTableNotFound(tableName string) error {	
+	return fmt.Errorf("%w %s", errors.New("Error: no such table: "), tableName)
```
tie, line 10: I believe the test fails due to this line—each error has a different memory location, so you can’t compare them directly.

sql/query/reindex.go
```diff
@@ -30,7 +30,8 @@ func (stmt ReIndexStmt) Run(ctx context.Context, tx *database.Transaction, args
 	if err == nil {
 		return res, t.ReIndex()
 	}
-	if err != database.ErrTableNotFound {
+	errReturned  := database.ErrTableNotFound(stmt.TableOrIndexName)
```
tie, line 33: In particular, this comparison should have been `errors.Is(err, database.ErrTableNotFound)`.

sql/query/drop.go
```diff
@@ -29,7 +29,8 @@ func (stmt DropTableStmt) Run(ctx context.Context, tx *database.Transaction, arg
 	}
 
 	err := tx.DropTable(stmt.TableName)
-	if err == database.ErrTableNotFound && stmt.IfExists {
+	errReturned := database.ErrTableNotFound(stmt.TableName)
+	if err == errReturned && stmt.IfExists {
```
tie, line 33: Use `errors.Is` here too.

database/errors.go
asdine: Very good idea!

tie:
@carlosfrodrigues thanks for the PR! I believe the tests fail because `ErrTableNotFound` function in your change returns a new error each time (both `fmt.Errorf` and `errors.New` return a new pointer). I’ve left a few suggestion in the review.

I also recommend reading [Working with Errors in Go 1.13](https://blog.golang.org/go1.13-errors) blog post.

database/errors.go
carlosfrodrigues: Thanks for the answer, I'm going to follow your suggestions :)

PR #249, opened 2020-10-10
RFC 001: Cancelable operations and use of Context

asdine:
This PR adds an RFC about how we should use `Context` in Genji.

It is opened for comments and suggestions.

### DIFF
--- a/rfcs/001_context.md
+++ b/rfcs/001_context.md
@@ -56,54 +56,25 @@ db, err := genji.New(ctx, ng)
 
 ### genji.DB methods
 
-Each of these must be cancelable independently, at any given time while the database handle is opened.
-Because cancelation is handled independently from the database handle, it is very intuitive and doesn't require preparing a different handle to be able to cancel a query.
+By default, database methods must not expect a context. This avoids context pollution when cancelation is not necessary.
+However, when it is necessary, a new handle can be created to use a specific context, using a new method named `WithContext`.
 
 ```go
-db.Exec(ctx, "INSERT INTO foo(a) VALUES (1)")
-db.Query(ctx, "SELECT 1")
-db.QueryDocument(ctx, "SELECT 1")
-```
+// non-cancelable queries
+db.Exec("INSERT INTO foo(a) VALUES (1)")
+db.Query("SELECT 1")
+db.QueryDocument("SELECT 1")
 
-Since a transaction is short-lived and atomic, contexts must be used to cancel the whole transaction by passing a context to `Begin`.
-Any call to the transaction handle methods must return an error and rollback the transaction if the context is canceled.
-
-```go
-tx, _ := db.Begin(ctx, true)
-defer tx.Rollback()
-
-tx.Exec("INSERT INTO foo(a) VALUES (1)")
-tx.Query("SELECT 1")
-tx.QueryDocument("SELECT 1")
-
-tx.Commit()
-```
-
-This also applies to `db.View` and `db.Update`:
-
-```go
-db.View(ctx, func(tx *Tx) error) error {
-    tx.Exec("INSERT INTO foo(a) VALUES (1)")
-    tx.Query("SELECT 1")
-    tx.QueryDocument("SELECT 1")
-})
-
-db.Update(ctx, func(tx *Tx) error) error {
-    tx.Exec("INSERT INTO foo(a) VALUES (1)")
-    tx.Query("SELECT 1")
-    tx.QueryDocument("SELECT 1")
-})
-```
-
-### The case of Close
-
-I don't think we should pass a context to `db.Close()`. Closing a database handle can take time, depending on the type of engine used and one might want to control that.
-But most of the time this is not important, so I'd rather add a specialized method.
-
-```go
 ctx, cancel := context.WithTimeout(context.Background(), 5 * time.Second)
 
-err = db.Shutdown(ctx)
+// WithContext returns a cancelable database handle
+db.WithContext(ctx).Exec("INSERT INTO foo (a) VALUES (1)")
+
+dbx := db.WithContext(ctx)
+
+// any called dbx method must return an error if the context is canceled.
+dbx.Query("SELECT 1")
+dbx.Begin(true)
 ```
 
 ## Engines
--- a/rfcs/001_context.md
+++ b/rfcs/001_context.md
@@ -14,7 +14,7 @@ The purpose of this RFC is not to describe in detail how each operation should b
 ### Open and New
 
 Opening a database relies on IO and should be cancelable, but after that operation is completed the database handle itself is long-lived and should not be cancelable, but closable with the `Close()` method.
-However, most of the time, we don't care about how long it takes.
+However, most of the time, we don't care about how long it takes to open a database.
 
 We currently have two ways of opening a database:
 
@@ -53,8 +53,6 @@ ng := memoryengine.New()
 // at startup, Genji performs some operations on the engine, this should be cancelable
 db, err := genji.New(ctx, ng)
 ```
-
-Storing the context for further uses prevents doing that and is not intuitive for a database handle that is supposed to be long-lived.
 
 ### genji.DB methods
 
@@ -118,6 +116,7 @@ ctx, cancel := context.WithTimeout(context.Background(), 5 * time.Second)
 // only the Begin method takes the context.
 tx, err := ng.Begin(ctx, true)
 
+// Transaction methods don't need to take a context.
 st, err := tx.GetStore([]byte("foo"))
 err = st.Put([]byte("a"), []byte("b"))
 
--- a/rfcs/001_context.md
+++ b/rfcs/001_context.md
@@ -81,6 +81,8 @@ dbx.Begin(true)
 
 Since almost every engine action is done from within a transaction, cancelation must be controlled by the context passed to the `Begin` method. Any action done on a canceled transaction must return an error.
 
+In addition, context cancellation or some unrecoverable I/O error may occur when iterating over store with `engine.Iterator`. In such case iterator is invalidated, `Valid()` method returns false and the new `Err()` will return non-nil error.
+
 ```go
 ctx, cancel := context.WithTimeout(context.Background(), 5 * time.Second)
 
@@ -93,6 +95,12 @@ err = st.Put([]byte("a"), []byte("b"))
 
 it := st.NewIterator(engine.IteratorConfig{})
 defer it.Close()
+for it.Seek(nil); it.Valid(); it.Next() {
+	// …
+}
+if err := it.Err(); err != nil {
+	return err
+}
 ```
 
 ### Packages that use an engine
--- a/rfcs/001_context.md
+++ b/rfcs/001_context.md
@@ -5,15 +5,15 @@
 
 ## General approach
 
-I would like to avoid context pollution as much as possible, but without restricting the user flexibility. It is important to be able to control and cancel operations but I also believe that there are times where we don't really care about that.
-I want Genji to be simple to use and intuitive, and for that I prefer to expand a bit more the surface API, rather than making the existing ones more complete and complex.
-The purpose of this RFC is not to describe in details how each individual operation should be cancelled but rather which APIs should expect a context parameter.
+I would like to avoid context pollution as much as possible, but without restricting user flexibility. It is important to be able to control and cancel operations but I also believe that there are times where we don't care about that.
+I want Genji to be simple to use and intuitive, and for that, I prefer to expand a bit more the surface API, rather than making the existing ones more complete and complex.
+The purpose of this RFC is not to describe in detail how each operation should be canceled but rather which APIs should expect a context parameter.
 
 ## Genji package and database lifecycle
 
 ### Open and New
 
-Opening a database relies on IO and should be cancelable, but after that operation is completed the database handle itself is long lived and should not be cancelable, but closable with the `Close()` method.
+Opening a database relies on IO and should be cancelable, but after that operation is completed the database handle itself is long-lived and should not be cancelable, but closable with the `Close()` method.
 However, most of the time, we don't care about how long it takes.
 
 We currently have two ways of opening a database:
@@ -34,13 +34,11 @@ if err != nil {
 defer db.Close()
 ```
 
-It is very simple to write and to use. I think it should remain as-is.
+It is very simple to write and to use, it must remain as-is.
 
-If users want more power, there is the `genji.New` function.
+If users want more power, they must use the `genji.New` function.
 
 Opening a database can take time, even for Bolt or Badger, because accessing disks is unbounded (slow disks, NFS, etc.). It makes sense to have this a cancelable operation.
-
-
 
 ```go
 ctx, cancel := context.WithTimeout(context.Background(), 5 * time.Second)
@@ -56,11 +54,11 @@ ng := memoryengine.New()
 db, err := genji.New(ctx, ng)
 ```
 
-Storing the context for further uses prevents doing that and is not really intuitive for a database handle that is supposed to be long lived.
+Storing the context for further uses prevents doing that and is not intuitive for a database handle that is supposed to be long-lived.
 
 ### genji.DB methods
 
-Each of these should be cancelable independently, at any given time while the database handle is opened.
+Each of these must be cancelable independently, at any given time while the database handle is opened.
 Because cancelation is handled independently from the database handle, it is very intuitive and doesn't require preparing a different handle to be able to cancel a query.
 
 ```go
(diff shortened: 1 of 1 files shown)


rfcs/001_context.md
```diff
@@ -0,0 +1,145 @@
...
+
+Storing the context for further uses prevents doing that and is not intuitive for a database handle that is supposed to be long-lived.
+
+### genji.DB methods
+
+Each of these must be cancelable independently, at any given time while the database handle is opened.
+Because cancelation is handled independently from the database handle, it is very intuitive and doesn't require preparing a different handle to be able to cancel a query.
+
+```go
+db.Exec(ctx, "INSERT INTO foo(a) VALUES (1)")
```
tie, line 65: I think we can reuse the context from `genji.Open`, and allow changing it with `WithContext` method. The shallow copy is cheap since `genji.DB` just wraps `database.Database`, same for the `genji.Tx`.

In fact, that’s what #231 already does (i.e. you only need explicit context on `Open`).

rfcs/001_context.md
```diff
@@ -0,0 +1,145 @@
...
+
+```go
+ctx, cancel := context.WithTimeout(context.Background(), 5 * time.Second)
+
+err = db.Shutdown(ctx)
+```
+
+## Engines
+
+Since almost every engine action is done from within a transaction, cancelation must be controlled by the context passed to the `Begin` method. Any action done on a canceled transaction must return an error.
```
tie, line 111: I think we need explicit context on engine transactions to address issues like #58.

rfcs/001_context.md
asdine: Reusing that context can cause several issues:

- It is less flexible: We can't control how long opening a database can take since the context is used for the whole database lifecycle, which is not supposed to be canceled very often.
- It is less intuitive: It is way more common to pass a context for the current operation rather than for the lifecycle of a database handle, which again is supposed to be long-lived.
- It makes the API more complex: We must remember to call `WithContext` every time we want one particular action to be cancelable, which is the most common use case. In the context of an http handler for example, we would have to always use `WithContext(req.Context())`. It seems easier to run `db.Exec(req.Context())` or `db.Begin(req.Context(), true)` instead.

If we want to automatically close the database when the root context is canceled, I believe there are simpler ways to do so

rfcs/001_context.md
asdine: Do you have some examples?

rfcs/001_context.md
tie: On the other hand, it’s more flexible because we can pass context to individual operations, and ignore it in case we don’t need it (e.g. command-line apps).

I don’t think it’s that that uncommon, a few examples I can think of right now are
- [http.Request.WithContext](https://golang.org/pkg/net/http#Request.WithContext)
- [gorm.DB.WithContext](https://pkg.go.dev/gorm.io/gorm#DB.WithContext)

It’s also nice to prototype some APIs without context cruft, and then gradually add contexts/cancellation where needed. We’ve been using `gorm` at my last $job, and I’d say that it’s less intrusive than `database/sql` interface.

Even if you missed the context, the change is one line:
```
tx = tx.WithContext(ctx)
// or
db = db.WithContext(ctx)
```

That said, I think that context-per-operation goes against the primary goal to avoid context pollution.

rfcs/001_context.md
tie: I’ll create a branch from #231 to show some examples once I review my changes.

rfcs/001_context.md
tie: One more objection for context per transaction is that we may want to set a timeout on an individual query, but not the whole transaction.

asdine: Ok so there are two subjects here:

1/ Whether we should reuse the context given during database opening
2/ Whether `db.Exec` and `db.Query*` should receive a context or not

My primary concern with 1/ is that the context should only be used for controlling database opening, and as proposed in this RFC, only passed to `genji.New`.

Regarding 2/ though, I think this is a good idea that can indeed prevent context pollution even further:

```go
db, _ := genji.Open("foo.db")
defer db.Close()

db.Exec("INSERT INTO foo (a) VALUES (1)")

db.WithContext(ctx).Exec("INSERT INTO foo (a) VALUES (1)")

dbx := db.WithContext(ctx)

dbx.Query("SELECT 1")

tx, err := dbx.Begin(true)

dbx.View(func() {...})
```

tie: >My primary concern with 1/ is that the context should only be used for controlling database opening, and as proposed in this RFC, only passed to genji.New.

We can address this by defaulting to `context.Background()` instead of the context passed to `Open`.

asdine: > We can address this by defaulting to context.Background() instead of the context passed to Open.

Exactly

asdine: > One more objection to the context per transaction is that we (as users) may want to set a timeout on an individual query, but not the whole transaction.

I thought about this but I have trouble seeing a valid use case that would be worth the trouble, i.e. when someone would need to cancel a specific query AND keep running other queries if that one was canceled.

tie:
I think `parser.ParseQuery` shouldn’t accept context (it does so now). Parsing statements shouldn’t be an expensive operation.

rfcs/001_context.md
```diff
@@ -0,0 +1,115 @@
...
+```go
+package database
+
+func (db *Database) Begin(ctx context.Context, writable bool) (*Transaction, error)
+func (db *Database) BeginTx(ctx context.Context, writable bool) (*Transaction, error)
+```
+
+Any function using the returned transaction must not expect a context in their signature.
+
+However, there might be some exceptions: If a function performs a long-running operation or relies on IO itself, it must expect a context.
```
yaziine, line 111: Do you have a concrete example for this?

asdine: > If a function performs a long-running operation or relies on IO itself

This basically means if a function does some other IO related work or some potentially long-running work (a very long for loop, reading from a file, etc.) that might need cancelation, it must expect a context.

yaziine: Why those kind of functions cannot use the context given to `Begin*`?

asdine: They are not supposed to hold the context just in case, they must explicitly expect it. It's much easier that way

asdine:
@tie @yaziine Do you see any blocker for this RFC?

rfcs/001_context.md
tie: Here is a commit that adds support for setting `(*bbolt.Bucket).FillPercent`: https://github.com/tie/genji/commit/3c8e2d86d90dde6aad1f0fb7df40d8bfd865b590

<details>
<summary>Examples</summary>


Set for DB initialization
```
var ctx context.Context // e.g. context.Background()
ctx = boltengine.FillPercentContext(ctx, 1.0)
genji.Open(ctx, ":memory:")
```

Set for all operations
```
var db *genji.DB
db.WithContext(boltengine.FillPercentContext(db.Context, 1.0))
```

Set for whole transaction (or a single operation)
```
var tx *genji.Tx
tx.WithContext(boltengine.FillPercentContext(tx.Context, 1.0))
```
</details>

I’d say that we should keep the API extensible for such cases, along with allowing cancelling operations in transaction.

Also, that makes it easier to reason about I/O usage in internal API. E.g. currently planner.Optimize needs a context because the rule UseIndexBasedOnSelectionNodeRule calls (*database.Table).Indexes which in turn calls I/O bound engine.Transaction.GetStore.
As you’ve said, only Bind needs a context, so this is an unwanted behavior.

rfcs/001_context.md
```diff
@@ -0,0 +1,115 @@
...
+// but not all of them
+ng := memoryengine.New()
+
+// at startup, Genji performs some operations on the engine, this should be cancelable
+db, err := genji.New(ctx, ng)
+```
+
+### genji.DB methods
+
+By default, database methods must not expect a context. This avoids context pollution when cancelation is not necessary.
```
tzzed, line 59: It is very useful that allows users to have flexibility in the methods usages. I love it.

yaziine (approved):
Everything seems fine to me. 👍

rfcs/001_context.md
asdine: I'm still not 100% sure we should do that. I can understand the flexibility that it would bring but I think it will lead to even more troubles.
In Postgres or SQLite for example, any canceled operation invalidates the entire transaction, whether it's a query error or a timeout for a specific statement.
And there are good reasons to do so: An error inside a transaction may lead to invalid / unrecoverable state, because an error can happen in the middle of internal writes. Delegating the error handling to the user would be useless because they would have to deal with duplicate errors, caches, etc.,. Example:

- `db.Foo()` does `write-A` + `write-B`
- `write-A` works
- `write-B` fails
- what should the user do?
- user calls `db.Foo()` again? `write-A` fails, already happened
- user calls `db.Bar()` instead? expects `write-A` and `write-B` but only finds `write-A`. Fails

Guaranteeing idempotency to the user in the middle of a transaction is too much work IMO, this seems much easier to simply cancel the transaction and let the user rollback.

tie: Thanks for explanation! I understand now that canceling operation in a transaction is not a good idea. I don’t think this neglects other benefits of having internal transaction API accept context though—specifically reasoning about I/O usage in internal API (e.g. planner.Optimize). That said, I agree that higher-level APIs should just reuse context from `BeginTx`.

asdine: Yes, any internal API can expect a context if they need to use it or to pass it. But, for this first iteration at least, they shouldn't care about what could happen if they call `engine.Store#Put` for example, as cancelation is handled by the transaction itself.

In other terms, if only `BeginTx` expects a transaction, most of the internal APIs that use the resulting `database.Transaction` type won't need to expect a context (unless they are doing some heavy work themselves).
This means for example that none of the `sql/*` packages need to expect a context (there might be some exceptions, but I believe we will see them during the implementation)

I'd rather do this iteratively by taking a cautious approach, then improve the internal API iteratively every time we think it makes sense to do so. What do you think?

asdine: @tie 🆙

tie: I understand your point, but I think it makes more sense to do the necessary *API* changes from the beginning and then incrementally reduce `context` usage (and thus IO usage) in routines like Optimize. For example, having database.Transaction (and underlying engine API) accept context could potentially save time on comments like https://github.com/genjidb/genji/pull/264#discussion_r511196836, both for author and reviewer.

Note that underlying engine transaction API would receive context set on *genji.Tx by (*genji.DB).Begin.

We can definitely split BeginTx and Get/Put/etc methods changes into two separate PRs to make review easier.

asdine: I'm probably missing the point but I still have trouble understanding why we need to pass the context to `tx.Foo()` if we will never use it. Could you give some examples of a specific API, how the context would be passed to it and how it would use that context?

asdine: > then incrementally reduce context usage (and thus IO usage) in routines like Optimize

By the way, this RFC reboots the usage of context everywhere. Once merged, we will remove all existing use of context that is not compliant with this proposal.

tie: The point is make API consistent in terms of context usage patterns. Under the current proposal both `genji` package and context-aware engine implementations would have to hold/manage a transaction context.

tie: A specific example would be tikv that needs a context for Commit method.
https://godoc.org/github.com/pingcap/tidb/kv#Transaction
https://github.com/pingcap/tidb/blob/974c5fa78e69af35879da1494b771f498ebfb1b4/kv/kv.go#L245-L246

asdine: > Under the current proposal root genji package and context-aware engine implementations would have to hold/manage a transaction context. We can offload this task to genji package and accept a context in engines via method arguments.

That's actually exactly what this proposal want, to delegate the management of the context to the `engine.Transaction` type.

If we don't do that, we are going to spread the context everywhere in the codebase for no actual gain. Every function that would need to interact with the `engine.Transaction` type would have to pass over the context 

<details>
<summary>engine/engine.go</summary>

```go
package engine

type Engine interface {
	Begin(ctx context.Context, writable bool) (Transaction, error)
	Close(ctx context.Context) error
}

type Transaction interface {
	Rollback(ctx context.Context) error
	Commit(ctx context.Context) error
	GetStore(ctx context.Context, name []byte) (Store, error)
	CreateStore(ctx context.Context, name []byte) error
	DropStore(ctx context.Context, name []byte) error
}

type Store interface {
	Get(ctx context.Context, k []byte) ([]byte, error)
	Put(ctx context.Context, k, v []byte) error
	Delete(ctx context.Context, k []byte) error
	Truncate(ctx context.Context) error
	NewIterator(IteratorConfig) Iterator
	NextSequence(ctx context.Context) (uint64, error)
} 

type Iterator interface {
	Seek(ctx context.Context, k []byte)
	Next(ctx context.Context)
	Valid() bool
	Item() Item
	Close(ctx context.Context) error
}
```
</details>


<details>
<summary>index/index.go</summary>

```go
package index


func (idx *Index) Set(ctx context.Context, v document.Value, k []byte) error {}
func (idx *Index) Delete(ctx context.Context, v document.Value, k []byte) error {}
func (idx *Index) AscendGreaterOrEqual(ctx context.Context, pivot document.Value, fn func(val, key []byte, isEqual bool) error) error {}

...
```
</details>

<details>
<summary>database/table.go</summary>

```go
package index


func (t *Table) Insert(ctx context.Context, d document.Document) ([]byte, error) {}
func (t *Table) Delete(ctx context.Context, key []byte) error {}
func (t *Table) Replace(ctx context.Context, key []byte, d document.Document) error {}
...
```
</details>

These are just a few examples that merely scratch the surface of APis that will need to change for the sole purpose of detecting IO cancelation in the engine.

This is exactly the same case as the new Go virtual file system API:

- We have one component whose implementations may or may not need to do IO (memoryengine doesn't do any IO, it does however perform long-running operations)
- Any of the methods of that component will be called within the lifetime of the same context

Now, here is a revised engine:

<details>
<summary>engine/engine.go</summary>

```go
package engine

type Engine interface {
    // we only need to pass the context here
	Begin(ctx context.Context, writable bool) (Transaction, error)
	Close(ctx context.Context) error
}

// all of these types are context aware
type Transaction interface {
	Rollback() error
	Commit() error
	GetStore(name []byte) (Store, error)
	CreateStore(name []byte) error
	DropStore(name []byte) error
}

type Store interface {
	Get(k []byte) ([]byte, error)
	Put(k, v []byte) error
	Delete(k []byte) error
	Truncate() error
	NewIterator(IteratorConfig) Iterator
	NextSequence() (uint64, error)
} 

type Iterator interface {
	Seek(k []byte)
	Next()
	Valid() bool
	Item() Item
	Close() error
}
```
</details>

- Only implementations need to deal with the context
- `engine/enginetest` will provide tests to make sure all implementations have a valid behavior
- Only `database.Begin*` needs to be modified

-> We have the same guarantees as before without changing much of the API.

Also, this solution is a much smaller step than spreading context everywhere just in case.
The primary goal of this RFC is to prevent unnecessary context pollution, and this solution seems to do the job.
We could observe how it goes and improve the solution progressively. Genji is still v0 and we can break APIs if needed.

tie: OK, no objections here then.

rfcs/001_context.md
```diff
@@ -0,0 +1,115 @@
...
+
+// any called dbx method must return an error if the context is canceled.
+dbx.Query("SELECT 1")
+dbx.Begin(true)
+```
+
+## Engines
+
+Since almost every engine action is done from within a transaction, cancelation must be controlled by the context passed to the `Begin` method. Any action done on a canceled transaction must return an error.
+
```
tie, line 83: One small suggestion though.

```suggestion

In addition, context cancellation or some I/O error may occur when iterating over store with `engine.Iterator`. In such case `Valid() bool` method returns false and the new `Err() error` will return non-nil error.

```

rfcs/001_context.md
```diff
@@ -0,0 +1,115 @@
...
+// only the Begin method takes the context.
+tx, err := ng.Begin(ctx, true)
+
+// Transaction methods don't need to take a context.
+st, err := tx.GetStore([]byte("foo"))
+err = st.Put([]byte("a"), []byte("b"))
+
+it := st.NewIterator(engine.IteratorConfig{})
+defer it.Close()
```
tie, line 95:
```suggestion
defer it.Close()
for it.Seek(nil); it.Valid(); it.Next() {
	// …
}
if err := it.Err(); err != nil {
	return err
}
```

tie (commented):
One last suggestion, everything else looks great!

rfcs/001_context.md
asdine: Good idea 👍🏼

Issue #43, opened 2019-10-20
Support for Auto increment


asdine:

89z:
https://github.com/genjidb/genji/pull/250

89z:
It looks like this is somewhat implemented:

https://github.com/genjidb/genji/blob/e55906e3f8559ae6db4ed57ecb12cd9f4f12d6ad/database/table.go#L43-L47

but I guess the generated key is not currently exposed?

89z:
It looks like it is currently supported, but Im not sure how you would go about
querying the ID:

~~~go
package main

import (
   "bytes"
   "github.com/genjidb/genji"
   "github.com/genjidb/genji/document"
   "log"
)

func check(e error) {
   if e != nil {
      log.Fatal(e)
   }
}

type Artist struct {
   ID int
   Name string
}

func main() {
   db, e := genji.Open("thunder.db")
   check(e)
   defer db.Close()
   e = db.Exec("CREATE TABLE artist_t")
   check(e)
   tx, e := db.DB.Begin(true)
   check(e)
   tb, e := tx.GetTable("artist_t")
   check(e)
   // 1
   d, e := document.NewFromStruct(Artist{Name: "Sting"})
   check(e)
   y1, e := tb.Insert(d)
   check(e)
   // 2
   d, e = document.NewFromStruct(Artist{Name: "Madonna"})
   check(e)
   y2, e := tb.Insert(d)
   check(e)
   // commit
   tx.Commit()
   println(bytes.Equal(y1, []byte{1}), bytes.Equal(y2, []byte{2}))
}
~~~

89z:
As a workaround, you can create a unique ID:

~~~go
func ID(d time.Duration) string {
   time.Sleep(d)
   n := time.Now().UnixNano() / int64(d)
   return big.NewInt(n).Text(62)
}
~~~

or you can get the current Max ID:

~~~go
func Max(db *genji.DB, table string) (float64, error) {
   d, e := db.QueryDocument("select max(id) from " + table)
   if e != nil {
      return 0, e
   }
   max, e := d.GetByField("max(id)")
   if e != nil {
      return 0, e
   }
   return max.V.(float64), nil
}
~~~

PR #250, opened 2020-10-10
Support AUTO_INCREMENT.

tzzed:
This a proposal draft for #43.

SQL Server seems good because it proposes a customizable `AUTO_INCREMENT` with a "natural language". 

```SQL 
### Default value
CREATE TABLE foo (id INTEGER AUTO_INCREMENT);
INSERT INTO foo VALUES {"a": "foo"};
SELECT * FROM foo;
{ "a": "foo", "id": 1}

### Set a start index and an increment value;
###`AUTO_INCREMENT(startIndex, incBy)`
###The first value of the sequence start at 10 and the next is incremented by 5
CREATE TABLE bar (id INTEGER AUTO_INCREMENT(10, 5));
INSERT INTO bar VALUES {"a": "bar"};
INSERT INTO bar VALUES {"a": "baz"};
SELECT * FROM bar;
{ "a": "bar", "id": 10 }
{  "a": "baz", "id": 15 }
```

`AUTO_INCREMENT` have to be applied only on number value type. `INTEGER` and `DOUBLE`
```
genji> CREATE TABLE foo(bar TEXT AUTO_INCREMENT);
genji> found text, expected integer, double at line 1, char 27
```

About `ALTER TABLE table_name AUTO_INCREMENT=100`, If we keep it like that, we should be able to write the both following syntaxes:

```SQL
###For default value
ALTER TABLE foo AUTO_INCREMENT=100;

### And this even if the creation was with default value
ALTER TABLE foo AUTO_INCREMENT(100, 10);
```

Thank you for your feedbacks.


tie:
@tzzed thanks for your work! What’s the current status of this PR? It looks like the counter is not being persisted.

<details>
<summary>Step 1</summary>

```
tie@xhyve ~/genji/cmd/genji $ go run . bolt.db
On-disk database using BoltDB engine at path bolt.db.
Enter ".help" for usage hints.
genji> CREATE TABLE foo (id INTEGER AUTO_INCREMENT);
genji> INSERT INTO foo VALUES {"a": "foo"};
genji> SELECT * FROM foo;
{
  "a": "foo",
  "id": 1
}
genji> .exit
```
</details>

<details>
<summary>Step 2</summary>

```
tie@xhyve ~/genji/cmd/genji $ go run . bolt.db
On-disk database using BoltDB engine at path bolt.db.
Enter ".help" for usage hints.
genji> SELECT * FROM foo;
{
  "a": "foo",
  "id": 1
}
genji> INSERT INTO foo VALUES {"a": "foo"};
genji> SELECT * FROM foo;
{
  "a": "foo",
  "id": 1
}
{
  "a": "foo",
  "id": 1
}
genji> INSERT INTO foo VALUES {"a": "foo"};
genji> SELECT * FROM foo;
{
  "a": "foo",
  "id": 1
}
{
  "a": "foo",
  "id": 1
}
{
  "a": "foo",
  "id": 2
}
genji> .exit
```
</details>

tzzed:
I just submitted the idea, before push the persistance. It was draft.
What do you think about the proposition?

tie:
Looks nice! I’d also suggest using `AUTOINCREMENT` as a keyword (underscore feels weird).

Though I’m not quite sure what we actually want from the auto increment feature. E.g. in SQLite it guarantees monotonically increasing 64-bit integer sequence for rowid (pk() in Genji) with an error on overflow. Also, I think it’s possible to compute non-primary key auto incremented value with (start, step) parameters given a document sequence number.

See also
- https://sqlite.org/autoinc.html
- https://sqlite.org/fileformat2.html#seqtab

tie:
Note that we currently don’t handle integer overflows for sequence numbers.

tie:
>If the AUTOINCREMENT keyword appears after INTEGER PRIMARY KEY, that changes the automatic ROWID assignment algorithm to prevent the reuse of ROWIDs over the lifetime of the database. In other words, the purpose of AUTOINCREMENT is to prevent the reuse of ROWIDs from previously deleted rows.
https://sqlite.org/autoinc.html

tzzed:
Thank you @tie ! In view of the documentation, this PR is more complex than I though. @asdine can you revert it. We should create a new table and more...
Thanks

mrusme:
Trying to understand whether auto incrementing primary keys work, even though this PR wasn't accepted. [The documentation](https://genji.dev/docs/genji-sql/table-basics/) says:

> This will create a table teams that can hold any document. An auto-incrementing primary key will be generated every time a document is inserted.

However, in a quick test locally this doesn't seem the case - at least not with only a `CREATE TABLE mytable;` schema. Do auto incrementing ID fields require a full schema specification?

asdine:
@mrusme For now, only auto generated "docids" are generated. These are internal IDs, equivalent to SQLite's rowids.
You can select them using `SELECT pk() FROM foo;` but like rowids you have no guarantee that they will always be the same.

Regarding auto incremented user defined primary keys, they are not supported yet.
I suggest using uuids or equivalent for now, until the feature is implemented.

Issue #252, opened 2020-10-11
panic: unknown operator "!~" / "=~"


tie:
`parser.ParseQuery` panics on the following input:
```
SELECT 0!~
SELECT 0=~
```

The culprits are `scanner.EQREGEX` and `scanner.NEQREGEX` (which are not currently supported).

<details>
<summary>cmd/genji stack trace</summary>

```
goroutine 1 [running]:
github.com/genjidb/genji/sql/parser.(*Parser).parseOperator(0xc00019c380, 0x16f3180, 0xc00019c3c0, 0x0, 0x0)
	github.com/genjidb/genji@v0.7.1/sql/parser/expr.go:136 +0x6ed
github.com/genjidb/genji/sql/parser.(*Parser).ParseExpr(0xc00019c380, 0x16f3180, 0xc00019c3c0, 0x0, 0x0, 0x0, 0x0)
	github.com/genjidb/genji@v0.7.1/sql/parser/expr.go:51 +0x22f
github.com/genjidb/genji/sql/parser.(*Parser).parseResultField(0xc00019c380, 0xc0001aa260, 0x6, 0x1166095, 0xc0001a2210)
	github.com/genjidb/genji@v0.7.1/sql/parser/select.go:91 +0x60
github.com/genjidb/genji/sql/parser.(*Parser).parseResultFields(0xc00019c380, 0x0, 0x0, 0x0, 0xc0001aa260, 0x6)
	github.com/genjidb/genji@v0.7.1/sql/parser/select.go:62 +0x2f
github.com/genjidb/genji/sql/parser.(*Parser).parseSelectStatement(0xc00019c380, 0x4f, 0x0, 0x0)
	github.com/genjidb/genji@v0.7.1/sql/parser/select.go:17 +0x71
github.com/genjidb/genji/sql/parser.(*Parser).ParseStatement(0xc00019c380, 0x4f, 0x0, 0x0, 0x0)
	github.com/genjidb/genji@v0.7.1/sql/parser/parser.go:68 +0x5ae
github.com/genjidb/genji/sql/parser.(*Parser).ParseQuery(0xc00019c380, 0xc00019c380, 0x80, 0xc0001a0680, 0x0, 0x203000)
	github.com/genjidb/genji@v0.7.1/sql/parser/parser.go:51 +0xde
[...]

PR #253, opened 2020-10-11
Added go-fuzz for parser

tie:
It’d be nice to integrate continuous fuzzing too, but I’m not aware of any easy to set up (and free) services.

See also #252


asdine:
Could we run the fuzzer on a Github Action when a branch is merged on master for example?

tie:
Yes, we can do that. I think it’s better to run a scheduled nightly fuzz test.

I’ll write a workflow that stores corpus in cache, but we’d probably need a separate repo or file store for `corpus` data if we don’t want to lose it.

asdine:
@tie Is this something that must be generated once and stored somewhere? If so, how about storing it in the Genji repo directly?

tie:
For each `Fuzz` function we need [corpus data](https://github.com/dvyukov/go-fuzz-corpus/tree/e6d670e/parser/corpus). Ideally, we should also update it since running go-fuzz may find new inputs that increase coverage (we can push from GitHub Actions workflow).

See https://github.com/google/fuzzing/blob/master/tutorial/libFuzzerTutorial.md#continuous-fuzzing

asdine:
@tie I have created a repo for storing the corpus data: https://github.com/genjidb/go-fuzz-corpus

.github/workflows/test.yaml
```diff
@@ -0,0 +1,57 @@
+name: Unit Tests
```
asdine, line 1: Thanks! Let's remove the Travis config then

fuzz/fuzz.go
```diff
@@ -0,0 +1,51 @@
...
+	t.Helper()
+
+	err := filepath.Walk("testdata", func(path string, info os.FileInfo, err error) error {
+		t.Helper()
+
+		if err != nil {
+			return err
+		}
+
+		if info.IsDir() && path == "suppressions" {
```
asdine, line 22: Could you add a comment here to explain why we are skipping this directory?

.github/workflows/test.yaml
tie: Note that `actions/setup-go` currently does not support gotip, though it should be easier to add now that they have the necessary infrastructure in [actions/go-versions](https://github.com/actions/go-versions).

fuzz/fuzz.go
tie: I’ve updated walkFn so that it only visits regular files in corpus and crashers directories.

asdine:
@tie Is it ready for review?

tie:
@asdine yes!

There would be another PR to `go-fuzz-corpus` repo with actions workflow to update the corpus on nightly schedule but the changes for main repo are ready. Here are the latest workflow runs on the PR source branch.
- [Fuzz Tests](https://github.com/tie/genji/runs/1253839914?check_suite_focus=true)
- [Unit Tests](https://github.com/tie/genji/runs/1253840265?check_suite_focus=true)
  Note that the (1.15.x, windows-latest, genji/engine/badgerengine) combo fails with `-race` flag. Should we (temporarily) exclude it from the matrix?

.github/workflows/test.yaml
tie: I’ve added codecov upload step in GitHub Actions, but I think we should keep Travis for `gotip` tests for now.

tie:
I’ve removed `go test` workflows since they are still work-in-progress. I’ll submit them as a new PR, and will also add `tinygo` test runner (currently we only run tests with `tinygo` tag in Travis config).

asdine (approved):
Thanks @tie ! That's great job! Let's iterate on this for tinygo and race on subsequent PRs 👍🏼

tie:
Neat, it’s working. Found another bug too!

PR #254, opened 2020-10-12
Add version subcommand to the CLI. Fixes #184

### LINKED ISSUE #184: Add the version command to the CLI
```bash
$ genji version
Genji v0.7.1
Genji CLI v0.7.1
```

This requires storing the Genji and Genji CLI version in a variable and building it the CLI using build flags.

```go
// somewhere in CLI code
const Version = "development"

// somewhere in Genji code
const Version = "development" 
```

On CI, whenever a tag is created, use build flags to override the relevant variable with the version information from the tag.

cvhariharan:
This PR adds a `version` subcommand which has a default value of `development` for both CLI and Genji. It automatically picks up the tag during CI.

Example output for `genji version`
```
Genji 1.0.0
Genji CLI 1.0.0
```
If tag is not set / empty, 
```
Genji development
Genji CLI development
```

### DIFF
--- a/cmd/genji/main.go
+++ b/cmd/genji/main.go
@@ -94,10 +94,10 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
 			Usage: "Shows Genji and Genji CLI version",
 			Action: func(c *cli.Context) error {
 				if CLIVersion == "" {
-					CLIVersion = "Development"
+					CLIVersion = "development"
 				}
 				if GenjiVersion == "" {
-					GenjiVersion = "Development"
+					GenjiVersion = "development"
 				}
 				fmt.Printf("Genji %v\nGenji CLI %v\n", GenjiVersion, CLIVersion)
 				return nil
--- a/cmd/genji/main.go
+++ b/cmd/genji/main.go
@@ -3,13 +3,11 @@ package main
 import (
 	"fmt"
 	"os"
+	"runtime/debug"
 
-	"github.com/genjidb/genji"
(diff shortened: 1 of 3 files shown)


cmd/genji/main.go
```diff
@@ -8,6 +8,9 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+var CLIVersion = "development"
+var GenjiVersion = "development"
```
asdine, line 12: I was thinking of storing this in Genji directly, and have the CLI use `genji.Version`. Would that work?

cvhariharan: I tried doing this but since we are actually building the cli, I am not able to pass the build flags to be set in genji. I will give this another look.

tie: Perhaps we can use [runtime/debug.ReadBuildInfo](https://golang.org/pkg/runtime/debug#ReadBuildInfo) to extract the main module version so that builds are easily reproducible?

tie: One thing to note about publishing build artifacts with Go modules is that instead of cloning the source we should `go get` tagged release (which should also add the new release to default $GOPROXY).

cvhariharan: @asdine I have moved `GenjiVersion` into `genji` package. If you feel like what @tie has suggested is a better approach, let me know. I will make the neccessary changes. :smile:

asdine: @cvhariharan @tie That's a great idea!

asdine (approved):
Thanks!

PR #260, opened 2020-10-14
NewFromStruct supports more types.

### LINKED ISSUE #225: Add support for embedded structs to document.NewFromStruct
```
type Embedded struct {
	Id string
}
type Example1 struct {
	*Embedded
}
type Example2 struct {
	Id string
}
```

`select * from examples where id = ?` will fail for the first struct but pass for the second struct - am i doing something wrong?

tzzed:
Fix #225 .

Use `FieldBuffer` instead of `StructDocument`.


document/create.go
```diff
@@ -75,90 +75,61 @@ func NewFromStruct(s interface{}) (Document, error) {
...
-		}
-
-		f := s.ref.Field(i)
+		sf := tp.Field(i)
 
-		v, err := NewValue(f.Interface())
-		if err != nil {
-			if err.(*ErrUnsupportedType) != nil {
-				continue
+		switch f.Kind() {
```
asdine, line 94: A switch with two cases is uncommon, how about having just an `if` with a `return` in it?

document/create.go
```diff
@@ -75,90 +75,61 @@ func NewFromStruct(s interface{}) (Document, error) {
...
+
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				fb.Add(strings.ToLower(sf.Name), NewDocumentValue(d))
+			}
+		default:
+			// Ignore unexported fields.
+			if sf.PkgPath == "" {
```
asdine, line 116: Try to always get rid of the edge case first:

```go
if sf.PkgPath != "" {
  return &fb, nil
}

...
```

https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

document/create.go
```diff
@@ -75,90 +75,61 @@ func NewFromStruct(s interface{}) (Document, error) {
...
-		return Value{}, ErrFieldNotFound
-	}
-
-	return NewValue(v.Interface())
-}
-
-// MarshalJSON implements the json.Marshaler interface.
-func (s structDocument) MarshalJSON() ([]byte, error) {
-	return jsonDocument{Document: s}.MarshalJSON()
+	return fb, nil
```
asdine, line 132:
```suggestion
	return &fb, nil
```

document/document_test.go
```diff
@@ -510,6 +537,51 @@ func TestNewFromStruct(t *testing.T) {
...
+			i   int // ignored
+			A   []byte
+		}
+
+		bb := baz{I: 100, A: "ptr"}
+		e := bar{Id: 1, Name: "Foo", Baz: bb}
+		f := foo{bar: &e, Bar: &e, B: "BString", A: []byte("foo"), i: 300}
+		doc, err := document.NewFromStruct(&f)
+		require.NoError(t, err)
+		fmt.Printf("docTest:: %v\n", document.NewDocumentValue(doc))
```
asdine, line 574:
```suggestion
```

document/document_test.go
```diff
@@ -658,7 +730,18 @@ func BenchmarkDocumentIterate(b *testing.B) {
 
 	b.Run("Reflection", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
-			refd, _ := document.NewFromStruct(&f)
+			refd, err := document.NewFromStruct(&f)
```
asdine, line 733: Since NewFromStruct returns now a FieldBuffer, we no longer need this test

asdine (commented):
Thanks @tzzed ! I have mostly Go related comments

document/create.go
tzzed: Yes with `continue` we just ignore the unexported field.

document/create.go
tzzed: There is no return inside the switch we are in a `for` loop. We can still do an `if-else` statement.

asdine: You can use `continue` instead or `return` in that case

document/create.go
```diff
@@ -75,94 +75,70 @@ func NewFromStruct(s interface{}) (Document, error) {
...
+		// Ignore unexported field.
+		if sf.PkgPath != "" {
+			continue
 		}
-		if strings.ToLower(sf.Name) == field {
-			ok = true
-			break
+
+		field := strings.ToLower(sf.Name)
+		if gtag, ok := sf.Tag.Lookup("genji"); ok && gtag != "-" {
```
yaziine, line 124: I am wondering if we should support the `omitempty` tag. What do you guys think? cc @asdine

yaziine: I think it can be useful if for example the user wants to add a field to the document only if it's not empty otherwise it is skipped.

```go
type Foo struct {
	Bar string
	Baz string `genji:"omitempty"`
}

foo1 := Foo{Bar:"bar", Baz:"baz"}
// {Bar:"bar", Baz:"baz"}

foo2 := Foo{Bar:"bar"}
// {Bar:"bar"} not {Bar:"bar", Baz:""}

```

tzzed: It make sense, just be sure is not a confusion point for the user. He should be sure that using `omitempty` disallow default value.

asdine: That's a good question, but since it's not directly related to this PR can you open an issue about this @yaziine?

document/create.go
```diff
@@ -75,94 +75,70 @@ func NewFromStruct(s interface{}) (Document, error) {
...
-// MarshalJSON implements the json.Marshaler interface.
-func (s structDocument) MarshalJSON() ([]byte, error) {
-	return jsonDocument{Document: s}.MarshalJSON()
+	return &fb, nil
 }
 
 // NewValue creates a value whose type is infered from x.
-func NewValue(x interface{}) (Value, error) {
+func
+NewValue(x interface{}) (Value, error) {
```
asdine, line 141: Could you explain this change? Are you using gofmt/goimports in your editor?

document/document_test.go
```diff
@@ -510,6 +536,50 @@ func TestNewFromStruct(t *testing.T) {
...
+		}
+
+		bb := baz{I: 100, A: "ptr"}
+		e := bar{Id: 1, Name: "Foo", Baz: bb}
+		f := foo{bar: &e, Bar: &e, B: "BString", A: []byte("foo"), i: 300}
+		doc, err := document.NewFromStruct(&f)
+		require.NoError(t, err)
+		v, err := doc.GetByField("id")
+		require.NoError(t, err)
+		require.EqualValues(t, 1, v.V.(int64))
```
asdine, line 575: `Id` has the `genji:"-"` tag, it should be ignored right?

document/document_test.go
```diff
@@ -357,10 +360,13 @@ func TestNewFromStruct(t *testing.T) {
 	q := 5
 	u.Q = &q
 	u.R = new(group)
+	u.S = u.R
```
asdine, line 363: Could you explain this change? It was originally to test the behavior with nil fields.

document/document_test.go
```diff
@@ -357,10 +360,13 @@ func TestNewFromStruct(t *testing.T) {
...
 	u.T = []int{1, 2, 3}
 	u.V = []*int{&q}
 	u.W = []user{u}
 	u.X = []interface{}{1, "foo"}
+	u.Y = [3]int{1, 2, 3}
```
asdine, line 368: Same here, could you explain why you are changing these?

document/create.go
tzzed: My bad, I am using both I just didn't run them. Corrected .

document/document_test.go
tzzed: The tests wasn't totally clear, I didn't know if it was wanted or not. Because some case was not handled.
I will change it.

document/document_test.go
tzzed: same as above.

document/document_test.go
tzzed: Yes I misunderstand, the field should be ignored.

document/document_test.go
asdine: I agree this test is not clear, I will add a few comments in the future

document/create.go
yaziine: FYI you can automatically run them on save. It's just a setting, just check with your editor.

tzzed: Done.

database/transaction_test.go
```diff
@@ -2,13 +2,13 @@ package database_test
...
 import (
 	"errors"
-	"testing"
 	"github.com/genjidb/genji/database"
 	"github.com/genjidb/genji/document"
 	"github.com/genjidb/genji/document/encoding/msgpack"
 	"github.com/genjidb/genji/engine/memoryengine"
 	"github.com/genjidb/genji/key"
 	"github.com/stretchr/testify/require"
+	"testing"
```
yaziine, line 11: This change shows that your editor is not well configured yet. `goimports` should have move this one with other built-in dependencies (e.g `errors`). 

`goimports` sorts the dependencies like this: 
- 1st block stdlib dependencies
- 2nd block the remaining deps

It can also separate your module dependency from the 3rd party dependencies in a 3rd block.

document/document_test.go
```diff
@@ -434,7 +439,7 @@ func TestNewFromStruct(t *testing.T) {
 			return nil
 		})
 		require.NoError(t, err)
-		require.Equal(t, 27, counter)
+		require.Equal(t, 25, counter)
```
yaziine, line 442: Could you explain this change? I am not sure to understand.

document/document_test.go
```diff
@@ -488,7 +493,18 @@ func TestNewFromStruct(t *testing.T) {
 		require.NoError(t, err)
 		d, ok := v.V.(document.Document)
 		require.True(t, ok)
-		v, err = d.GetByField("a")
+		v, err = d.GetByField("gp")
+		require.NoError(t, err)
+		require.EqualValues(t, 0, v.V.(int64))
+
+		v, err = doc.GetByField("q")
```
yaziine, line 500: Thanks for improving the tests! 🎉

document/document_test.go
```diff
@@ -510,6 +526,49 @@ func TestNewFromStruct(t *testing.T) {
 		require.NoError(t, err)
 		require.EqualValues(t, 2, v.V.(int64))
 	})
+
+	t.Run("Embedded struct", func(t *testing.T) {
+		db, err := genji.Open(":memory:")
```
yaziine, line 531: I think that we can avoid using a real database for testing this. What about using the `GetByField` function just above?

database/transaction_test.go
tzzed: Now my IDE is configured.

document/document_test.go
tzzed: The previous review, I added some tests  and in this one I delete them. See previous review.

document/document_test.go
tzzed: `GetByField` is used https://github.com/genjidb/genji/blob/47bae6e2b7a6c1de9277e3114a5a3038879806b0/document/document_test.go#L563

I did it because at first when I wrote my tests, I did used the issue example after some work I deleted queries `Insert` and `Select`... and I forgave to delete the `create` statement and `open`.
Indeed we should work only with the returned `document` and its methods.
Going to delete.

document/create.go
tzzed: Could I mark this as resolve ?

document/document_test.go
```diff
@@ -402,29 +405,31 @@ func TestNewFromStruct(t *testing.T) {
 			case 14:
 				require.EqualValues(t, document.DocumentValue, v.Type)
 			case 15:
-				require.EqualValues(t, document.NullValue, v.Type)
```
asdine, line 405: Could you explain why this has changed? This means that empty pointers are transformed to integers?

database/transaction_test.go
asdine: @tzzed It must have been done because your editor wasn't properly configured. Now that it is properly setup, please correct this file because as you can see, it is modified for no reason and the modification is wrong because the order is not the right one.

sql/parser/expr.go
```diff
@@ -309,15 +309,15 @@ func (p *Parser) parseType() (document.ValueType, error) {
 		p.Unscan()
 		return document.DoubleValue, nil
 	case scanner.TYPEINTEGER, scanner.TYPEINT, scanner.TYPEINT2, scanner.TYPEINT8, scanner.TYPETINYINT,
-		 scanner.TYPEBIGINT, scanner.TYPEMEDIUMINT, scanner.TYPESMALLINT:
+		scanner.TYPEBIGINT, scanner.TYPEMEDIUMINT, scanner.TYPESMALLINT:
```
asdine, line 312: Please remove this file from this PR as it is not directly related

sql/query/reindex.go
```diff
@@ -3,7 +3,7 @@ package query
 import (
 	"context"
 	"errors"
-	
+
```
asdine, line 6: Please remove this file from this PR as it is not directly related

document/document_test.go
```diff
@@ -286,7 +290,7 @@ func TestNewFromMap(t *testing.T) {
 
 func TestNewFromStruct(t *testing.T) {
```
asdine, line 291: This test is here to indicate if the `NewFromStruct` implementation is correct or not, it should not be modified for no reason. I can see from the modifications you wrote below that the new implementation is not correct anymore.
We should undo the modifications below (except the one directly related to the embed struct of course) and modify `NewFromStruct` until it passes all the tests

document/document_test.go
```diff
@@ -510,6 +510,51 @@ func TestNewFromStruct(t *testing.T) {
...
+		}
+
+		type foo struct {
+			*bar
+			B   string `genji:"b-b"`
+			Bar bar
+			i   int  // ignored
+			P   *int // should be null
+			A   []byte
+		}
```
yaziine, line 532: Instead of declaring new types why don't we use the one that already exists? The `user` struct with the embeded `group`.

tzzed: I prefered wrote a `struct` that I "master"  with my own tests due to the complexity of the the `struct user`.

yaziine (approved):
Thanks @tzzed, everything seems fine to me now! 🎉

document/document_test.go
```diff
@@ -656,12 +665,4 @@ func BenchmarkDocumentIterate(b *testing.B) {
...
 	})
 
-	b.Run("Reflection", func(b *testing.B) {
-		for i := 0; i < b.N; i++ {
-			refd, _ := document.NewFromStruct(&f)
-			refd.Iterate(func(string, document.Value) error {
-				return nil
-			})
-		}
-	})
```
tie, line 666: Could you explain why this benchmark was removed?

tie (commented):
Thanks! A have a few general questions (not suggestions per se) concerning the changes.

- Perhaps we should be rename `NewFromStruct` to `NewFieldBufferFromStruct` since it’s now building the `FieldBuffer`?
- Why can’t we keep the `structDocument` and descend into nested structs / anonymous fields on iteration?

document/document_test.go
tzzed: It because now we are using a `FieldBuffer`, @asdine told me to delete it for this reason.

> * Perhaps we should be rename `NewFromStruct` to `NewFieldBufferFromStruct` since it’s now building the `FieldBuffer`?
 `NewFromStruct` returns a `Document`. 


> * Why can’t we keep the `structDocument` and descend into nested structs / anonymous fields on iteration?
I think using `FieldBuffer` is a good way because it already implements `GetByField` and `Iterate`.

tie: > It because now we are using a FieldBuffer

Ah, that’s right.

>I think using FieldBuffer is a good way because it already implements GetByField and Iterate.

I’m fine with both approaches, though I definitely have a preference for the laziness of `structDocument`.

asdine: I have the same preference, but since `NewFromStruct` is mostly used when inserting data, I don't have strong reasons to refuse this change.

asdine:
@tzzed Could you rebase this branch please?

asdine (approved):
Thanks @tzzed !

Issue #261, opened 2020-10-15
Panic on genji version


tdemin:
After having run `go get -u -v github.com/genjidb/genji/cmd/genji`, trying to run `genji version` results in this output:

```
~ > genji version
Using z.Allocator with starting ref: c74a000000000000
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x919018]

goroutine 1 [running]:
main.main.func2(0xc00007cb80, 0x1, 0x1)
        /home/tdemin/go/src/github.com/genjidb/genji/cmd/genji/main.go:100 +0x58
github.com/urfave/cli.(*Command).Run(0xc0000f6d80, 0xc00007cac0, 0x0, 0x0)
        /home/tdemin/go/src/github.com/urfave/cli/command.go:163 +0x4ed
github.com/urfave/cli.(*App).RunContext(0xc000001980, 0xb69e00, 0xc0000281f0, 0xc00000e080, 0x2, 0x2, 0x0, 0x0)
        /home/tdemin/go/src/github.com/urfave/cli/app.go:306 +0x81f
github.com/urfave/cli.(*App).Run(...)
        /home/tdemin/go/src/github.com/urfave/cli/app.go:217
main.main()
        /home/tdemin/go/src/github.com/genjidb/genji/cmd/genji/main.go:143 +0x612
```

tdemin:
The Git commit I built Genji from is 375ed7aafa0f1d508d14671adb450ac6eb4789dd.

asdine:
I supposed you run `go get` outside of a module? When that happens `debug.ReadBuildInfo` returns nil, which then panics here: https://github.com/genjidb/genji/blob/375ed7aafa0f1d508d14671adb450ac6eb4789dd/cmd/genji/main.go#L100

tie:
I think we should also provide official binaries via GitHub releases. The GOPATH `go get` is essentially development version, so we can’t do much about it.

asdine:
Closed by #262 

@tie good idea, I added an issue about that here https://github.com/genjidb/genji/issues/269

PR #262, opened 2020-10-15
Do not panic on genji version when compiled in GOPATH mode

tdemin:
The `genji version` command relies on the versions of modules compiled in as debug info. Those are unavailable when compiling Genji with `go get github.com/genjidb/genji/cmd/shell` (in GOPATH mode). This makes `genji version` display a message saying the version is not available if genji has been built with `go get`.

Sample output:

```
genji on master [!?] via v1.15.2 > GO111MODULE=off go run github.com/genjidb/genji/cmd/genji version
Using z.Allocator with starting ref: 8005000000000000
version not available in GOPATH mode
```

### DIFF
--- a/cmd/genji/main.go
+++ b/cmd/genji/main.go
@@ -93,10 +93,13 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
 			Action: func(c *cli.Context) error {
 				var cliVersion, genjiVersion string
 				info, ok := debug.ReadBuildInfo()
-				if ok {
-					cliVersion = info.Main.Version
+
+				if !ok {
+					fmt.Println("version not available in GOPATH mode")
+					return nil
 				}
 
+				cliVersion = info.Main.Version
 				for _, mod := range info.Deps {
 					if mod.Path != "github.com/genjidb/genji" {
 						continue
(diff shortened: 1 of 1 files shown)


tdemin:
I was thinking of returning an additional error code with `cli.NewExitError("...", 1)`, but this would add more visual noise (it's already 2 lines of output, and this would add the 3rd one with an exit code).

tdemin:
Related issue: #261

cmd/genji/main.go
```diff
@@ -93,10 +93,13 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
 			Action: func(c *cli.Context) error {
 				var cliVersion, genjiVersion string
 				info, ok := debug.ReadBuildInfo()
-				if ok {
-					cliVersion = info.Main.Version
+
+				if !ok {
+					fmt.Println("version not available in GOPATH mode")
```
tie, line 98: Perhaps we should suggest running `go get` with modules enabled? Something like
```
fmt.Println("version not available in GOPATH mode; use \"go get\" with Go modules enabled")
```

tie:
@tdemin thanks for this change, nice catch! I’ve added a minor suggestion for the GOPATH mode message.

tdemin:
@tie sure thing, just added a notice on using modules.

tie (approved):

asdine (approved):
Thanks @tdemin !

PR #264, opened 2020-10-15
Add SELECT DISTINCT queries support

tdakkota:
Adds new `DISTINCT` keyword 
Uses hash set to make projection result unique.

### DIFF
--- a/document/document.go
+++ b/document/document.go
@@ -75,34 +75,6 @@ func Fields(d Document) ([]string, error) {
 	return fields, nil
 }
 
-func IterateInOrder(d Document, fn func(string, Value) error) error {
-	type pair struct {
-		field string
-		value Value
-	}
-	var pairs []pair
-
-	err := d.Iterate(func(field string, value Value) error {
-		pairs = append(pairs, pair{field, value})
-		return nil
-	})
-	if err != nil {
-		return err
-	}
-
-	sort.Slice(pairs, func(i, j int) bool {
-		return strings.Compare(pairs[i].field, pairs[j].field) == -1
-	})
-
-	for _, p := range pairs {
-		err := fn(p.field, p.value)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
 // FieldBuffer stores a group of fields in memory. It implements the Document interface.
 type FieldBuffer struct {
 	fields []fieldValue
--- a/document/document.go
+++ b/document/document.go
@@ -75,6 +75,34 @@ func Fields(d Document) ([]string, error) {
 	return fields, nil
 }
 
+func IterateInOrder(d Document, fn func(string, Value) error) error {
+	type pair struct {
+		field string
+		value Value
+	}
+	var pairs []pair
+
+	err := d.Iterate(func(field string, value Value) error {
+		pairs = append(pairs, pair{field, value})
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	sort.Slice(pairs, func(i, j int) bool {
+		return strings.Compare(pairs[i].field, pairs[j].field) == -1
+	})
+
+	for _, p := range pairs {
+		err := fn(p.field, p.value)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // FieldBuffer stores a group of fields in memory. It implements the Document interface.
 type FieldBuffer struct {
 	fields []fieldValue
--- a/sql/planner/distinct.go
+++ b/sql/planner/distinct.go
@@ -1,0 +1,33 @@
+package planner
+
+import (
+	"github.com/genjidb/genji/database"
+	"github.com/genjidb/genji/document"
+	"github.com/genjidb/genji/sql/query/expr"
+)
+
+type dedupNode struct {
+	node
+}
+
+func NewDistinctNode(n Node) Node {
+	return &dedupNode{
+		node{
+			op:   Dedup,
+			left: n,
+		},
+	}
+}
+
+func (n *dedupNode) Bind(tx *database.Transaction, params []expr.Param) error {
+	return nil
+}
+
+func (n *dedupNode) toStream(st document.Stream) (document.Stream, error) {
+	set := newDocumentHashSet(nil) // use default hashing algorithm
+	return st.Filter(set.Filter), nil
+}
+
+func (n *dedupNode) String() string {
+	return "Dedup()"
+}
--- a/sql/planner/distinct.go
+++ b/sql/planner/distinct.go
@@ -13,7 +13,7 @@ type dedupNode struct {
 	indexes   map[string]database.Index
 }
 
-func NewDistinctNode(n Node, tableName string) Node {
+func NewDedupNode(n Node, tableName string) Node {
 	return &dedupNode{
 		node: node{
 			op:   Dedup,
--- a/sql/planner/distinct.go
+++ b/sql/planner/distinct.go
@@ -8,19 +8,29 @@ import (
 
 type dedupNode struct {
 	node
+
+	tableName string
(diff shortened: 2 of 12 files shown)


sql/planner/projection.go
```diff
@@ -37,6 +38,18 @@ func NewProjectionNode(n Node, expressions []ProjectedField, tableName string) N
 	}
 }
 
+func NewDistinctProjectionNode(n Node, expressions []ProjectedField, tableName string) Node {
```
asdine, line 41: Distinct should be its own node, all it does is filtering duplicate documents with the same shape.
We could add a `Dedup` operation, that would deduplicate any document from a stream, using Filter with a document Hashset.

Here is how it would look like and where it would be inserted in the final stream:

```sql
genji> EXPLAIN SELECT a, b FROM foo WHERE c > 10 ORDER BY d LIMIT 1 OFFSET 2;
{
  "plan": "Table(foo) -> σ(cond: c > 10) -> ∏(a, b) -> Sort(d ASC) -> Offset(2) -> Limit(1)"
}

genji> EXPLAIN SELECT DISTINCT a, b FROM foo WHERE c > 10 ORDER BY d LIMIT 1 OFFSET 2;
{
  "plan": "Table(foo) -> σ(cond: c > 10) -> ∏(a, b) -> Dedup() -> Sort(d ASC) -> Offset(2) -> Limit(1)"
}
```

sql/query/select_test.go
```diff
@@ -222,3 +224,43 @@ func TestSelectStmt(t *testing.T) {
...
+
+	db, err := genji.Open(":memory:")
+	require.NoError(t, err)
+	defer db.Close()
+
+	err = db.Exec(ctx, "CREATE TABLE test(a integer)")
+	require.NoError(t, err)
+
+	for i := 0; i < 100; i++ {
+		err = db.Exec(ctx, `INSERT INTO test VALUES {a: ?, b: ?, c: {a: ?, b: ?}}`, i%10, i, i%10, i%10+1)
```
asdine, line 239: We might want to test the behavior with different types, especially NULL values

asdine (commented):
Thanks!

sql/planner/hash_set.go
```diff
@@ -0,0 +1,65 @@
...
+
+		buf = buf[0:]
+		return nil
+	})
+	if err != nil {
+		return 0, err
+	}
+
+	return s.hash.Sum64(), nil
+}
```
tdakkota, line 50: This function relies on `Iterate` function implementation, so if it does not guarantee order of field iteration (for example `map`-based) then hash will be incorrect.
For now, Genji does not have such implementation, but it can be added later.
I think `document.Value`/`document.Document` should have some `Hash` function to use it in hash table.

sql/planner/hash_set.go
```diff
@@ -39,10 +39,14 @@ func (s documentHashSet) generateKey(d document.Document) (uint64, error) {
 			return err
 		}
 
+		buf = buf[0:]
```
tie, line 42: This seems redundant, unless `var buf []byte` is declared outside of the `Iterate` function.

tie: Also, should be `buf[:0]` to reset slice length.

sql/planner/optimizer.go
```diff
@@ -248,6 +249,78 @@ func RemoveUnnecessarySelectionNodesRule(t *Tree) (*Tree, error) {
...
+			if !ok {
+				continue
+			}
+
+			pn, ok := d.left.(*ProjectionNode)
+			if !ok {
+				continue
+			}
+
+			table, err := pn.tx.GetTable(pn.tableName)
```
tie, line 270: Can we get this information on `Bind` instead?

tie (commented):
Thanks! I’ve some comments after skimming through the changes.

tie:
@tdakkota thanks! Can you please mark the discussions as resolved? It looks like only PR author can do that in GitHub…

sql/planner/distinct.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type dedupNode struct {
+	node
+
+	tableName string
+	indexes   map[string]database.Index
+}
+
+func NewDistinctNode(n Node, tableName string) Node {
```
asdine, line 16: We should rename this, Dedup is the name of an operation that could be used in several cases, DISTINCT is the name of a specific SQL feature.

```suggestion
func NewDedupNode(n Node, tableName string) Node {
```

sql/planner/optimizer.go
```diff
@@ -248,6 +249,68 @@ func RemoveUnnecessarySelectionNodesRule(t *Tree) (*Tree, error) {
 	return t, nil
 }
 
+// RemoveUnnecessaryDedupNodeRule removes any Dedup nodes
+// where projection is already unique.
+func RemoveUnnecessaryDedupNodeRule(t *Tree) (*Tree, error) {
```
asdine, line 254: 👏🏼

sql/planner/tree.go
```diff
@@ -43,6 +43,9 @@ const (
 	// Unset is an operation that removes a path from every document of a stream
 	Unset
 	// Group is an operation that groups documents based on a given path.
+	_
+	// Dedup
```
asdine, line 47:
```suggestion
	// Dedup is an operation that removes duplicate documents from a stream
```

sql/planner/hash_set.go
```diff
@@ -0,0 +1,64 @@
...
+		set:  map[uint64]struct{}{},
+	}
+}
+
+func (s documentHashSet) generateKey(d document.Document) (uint64, error) {
+	defer s.hash.Reset()
+
+	err := d.Iterate(func(field string, value document.Value) error {
+		var buf []byte
+		buf, err := key.AppendValue(buf, value)
```
tie, line 32:
```suggestion
		buf, err := key.AppendValue(nil, value)
```

sql/planner/hash_set.go
asdine: We should not rely on the implicit behavior of Document implementations as the `Document` interface doesn't guarantee the order of iteration.
Instead, we must use the [document.Fields()](https://github.com/genjidb/genji/blob/c05715e76a2436868192fc3ac47dc9e8c18e3437/document/document.go#L60) function, which gets the document keys and sorts them, and use `GetByField` for each key.

tdakkota: `pk()`/`literal` tests was failing, when I used `documentMask.GetByField`, it tries to get a value from masked document, but literals and functions values are not stored in document.
So, I added  `IterateInOrder` function.

document/document.go
```diff
@@ -75,6 +75,34 @@ func Fields(d Document) ([]string, error) {
 	return fields, nil
 }
 
+func IterateInOrder(d Document, fn func(string, Value) error) error {
```
asdine, line 78: I'm not sure this is necessary, as 60% of this function is already achieved by `document.FieldBuffer` and the `document.Fields` function:

```go
fields, _ := document.Fields(d)
for _, field := range fields {
  v, _ := d.GetByField(field)
}
``` 

Also, `document.Fields` automatically detect if the document is a `FieldBuffer` and take the short path by calling `FieldBuffer.Fields()`

sql/planner/hash_set.go
asdine: It appears to be a bug, they should be stored in `DocumentMask`:

```go
genji> create table foo;
genji> insert into foo(a) values (1), (2), (3);
genji> select pk() from foo;
{
  "pk()": 1
}
{
  "pk()": 2
}
{
  "pk()": 3
}
```

asdine (approved):
Thanks @tdakkota !

PR #270, opened 2020-10-17
Slightly more efficient conversion from JSON to document

asdine:
This PR is a small, low hanging, premature optimization.
It changes the way the `document.NewFromJSON` function works by lazily decoding the json data and delegating error handling to the `Iterate` and `GetByField` methods.
This takes advantage of the `jsonparser` library and prevents from duplicating the entire json object in memory before storing it.

It's 50% faster for small objects and about 20% for large objects, and 80% more memory efficient. It might be useful later for quick stream insertion.

```
name               old time/op    new time/op    delta
JSONToDocument-12    9.17µs ± 0%    7.86µs ± 0%   ~     (p=1.000 n=1+1)

name               old alloc/op   new alloc/op   delta
JSONToDocument-12    5.65kB ± 0%    3.10kB ± 0%   ~     (p=1.000 n=1+1)

name               old allocs/op  new allocs/op  delta
JSONToDocument-12       106 ± 0%       100 ± 0%   ~     (p=1.000 n=1+1)
```

### DIFF
--- a/document/create.go
+++ b/document/create.go
@@ -9,16 +9,40 @@ import (
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/buger/jsonparser"
 )
 
-// NewFromJSON creates a document from a JSON object.
-func NewFromJSON(data []byte) (Document, error) {
-	var fb FieldBuffer
-	err := fb.UnmarshalJSON(data)
+// NewFromJSON creates a document from raw JSON data.
+// The returned document will lazily decode the data.
+// If data is not a valid json object, calls to Iterate of GetByField will
+// return an error.
+func NewFromJSON(data []byte) Document {
+	return &jsonEncodedDocument{data}
+}
+
+type jsonEncodedDocument struct {
+	data []byte
+}
+
+func (j jsonEncodedDocument) Iterate(fn func(field string, value Value) error) error {
+	return jsonparser.ObjectEach(j.data, func(key, value []byte, dataType jsonparser.ValueType, offset int) error {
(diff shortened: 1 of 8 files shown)


document/create.go
```diff
@@ -9,16 +9,43 @@ import (
...
+	"github.com/buger/jsonparser"
 )
 
-// NewFromJSON creates a document from a JSON object.
-func NewFromJSON(data []byte) (Document, error) {
-	var fb FieldBuffer
-	err := fb.UnmarshalJSON(data)
+// NewFromJSON creates a document from raw JSON data.
+// The returned document will lazily decode the data.
+// If data is not a valid json object, calls to Iterate of GetByField will
```
asdine, line 18:
```suggestion
// If data is not a valid json object, calls to Iterate or GetByField will
```

tie:
It looks like you din’t set the `-count` flag when running benchmarks (`p=1.000 n=1+1`).

asdine:
10 times 👍🏼 
```
name               old time/op    new time/op    delta
JSONToDocument-12    8.39µs ± 3%    7.06µs ± 2%  -15.82%  (p=0.000 n=10+9)

name               old alloc/op   new alloc/op   delta
JSONToDocument-12    5.65kB ± 0%    3.10kB ± 0%  -45.04%  (p=0.000 n=10+10)

name               old allocs/op  new allocs/op  delta
JSONToDocument-12       106 ± 0%       100 ± 0%   -5.66%  (p=0.000 n=10+10)
```

PR #276, opened 2020-10-18
Fix fuzz workflow failing when there are no crashers

tie:
Interrupting go-fuzz always returns non-zero error code, so we should ignore it. Check whether `crashers` directory exists instead.


asdine (approved):
Good idea!

PR #279, opened 2020-10-21
Add constraint checking when creating table

### LINKED ISSUE #222: Prevent creating a table if constraints are incoherent
```sql
CREATE TABLE foo (
  a INTEGER,
  a.b[0] TEXT
);
```

In this example, `a` is expected to be both an integer an a document, which is not possible. Detect these kind of incoherence and return an error.

tdakkota:
Fixes #222.


sql/parser/constraint_tree.go
```diff
@@ -0,0 +1,116 @@
...
+func (n *constraintNode) getPath() document.ValuePath {
+	if n.parent == nil {
+		return document.ValuePath{n.frag}
+	}
+
+	return append(n.parent.getPath(), n.frag)
+}
+
+func (n *constraintNode) search(path document.ValuePath) *constraintNode {
+	if n.frag == path[0] {
```
tie, line 49: Keep the happy path unindented.
```
if n.frag != path[0] {
	return nil
}
```

tie (commented):
Wow, awesome work! 🎉
I only have some minor style comments and concerning `constraint_tree.go` file name—underscores are usually used for Go build constraints (either implicit or to indicate an explicit constraint) so perhaps something like `constraint.go` would be a better name.

tie:
I think we can use the tree-like field constraint representation internally too in the next iteration.

```
genji> CREATE TABLE t (a INTEGER, a.b[0] TEXT);
incoherent field constraint: "a.b" already exists as type a, but trying add "integer" constraint
```

```
genji> CREATE TABLE t (a INTEGER);
genji> ALTER TABLE t ADD FIELD a.b[0] TEXT;
```

sql/parser/constraint_tree.go
tzzed, line 49: I think you should check the non equal `path` first for more readability.
What do you think about that? 
```go
    if n.frag != path[0] {
          return nil
   }
  ....
```

sql/parser/constraint_tree.go
```diff
@@ -0,0 +1,116 @@
...
+		if len(path) == 1 {
+			return n
+		}
+
+		for _, sub := range n.sub {
+			t := sub.search(path[1:])
+			if t != nil {
```
tzzed, line 56: You can reduce one more here.  
```go
        if t := sub.search(path[1:]); t != nil {
                 return t
        }
```

tzzed (commented):
Very good job @tdakkota I just left a few comments.

sql/parser/create.go
```diff
@@ -108,6 +107,11 @@ func (p *Parser) parseFieldConstraints(info *database.TableInfo) error {
 			return err
 		}
 
+		err := tree.insert(fc.Path, fc.Type)
```
tzzed, line 110: You can reduce this line.
```go
          	if err := tree.insert(fc.Path, fc.Type); err != nil {
			return fmt.Errorf("incoherent field constraint: %w", err)
		}
```

sql/parser/create.go
```diff
@@ -108,6 +107,11 @@ func (p *Parser) parseFieldConstraints(info *database.TableInfo) error {
...
+		if err != nil {
+			return fmt.Errorf("incoherent field constraint: %w", err)
+		}
```
tzzed, line 113: You can reduce this line.
```go
          	if err := tree.insert(fc.Path, fc.Type); err != nil {
			return fmt.Errorf("incoherent field constraint: %w", err)
		}
```

sql/parser/constraint.go
```diff
@@ -0,0 +1,118 @@
+package parser
+
+import (
+	"fmt"
+
+	"github.com/genjidb/genji/document"
+)
+
+type constraintNode struct {
```
asdine, line 9: A small comment describing what this is used for would be perfect

asdine (commented):
Thanks @tdakkota ! This is indeed great work!
I think the logic detecting incoherent table constraints should be called from within [query.CreateStmt](https://github.com/genjidb/genji/blob/d4d3e0709cdc92f35661530f05611e81dbb38668/sql/query/create.go#L26-L39) type, rather than directly from the parser.

Ideally, the parser should only deal with grammar issues and delegate more complex error handling to the statement types.

asdine (approved):
Thanks @tdakkota !

PR #280, opened 2020-10-21
Add trailing .0 for double value when needed

### LINKED ISSUE #237: Display DOUBLES  with a decimal point

yaziine:
This fixes #237.

### DIFF
--- a/document/value_test.go
+++ b/document/value_test.go
@@ -21,7 +21,7 @@ func TestValueString(t *testing.T) {
 		{"bool", document.NewBoolValue(true), "true"},
 		{"int", document.NewIntegerValue(10), "10"},
 		{"double", document.NewDoubleValue(10.1), "10.1"},
-		{"double with trailing .0", document.NewDoubleValue(10), "10.0"},
+		{"double with trailing .0", document.NewDoubleValue(9223372036854775808), "9223372036854775808.0"},
 		{"document", document.NewDocumentValue(document.NewFieldBuffer().Add("a", document.NewIntegerValue(10))), "{\"a\": 10}"},
 		{"array", document.NewArrayValue(document.NewValueBuffer(document.NewIntegerValue(10))), "[10]"},
 	}
--- a/document/value_test.go
+++ b/document/value_test.go
@@ -21,6 +21,7 @@ func TestValueString(t *testing.T) {
 		{"bool", document.NewBoolValue(true), "true"},
 		{"int", document.NewIntegerValue(10), "10"},
 		{"double", document.NewDoubleValue(10.1), "10.1"},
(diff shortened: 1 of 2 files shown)


document/value_test.go
```diff
@@ -21,6 +21,7 @@ func TestValueString(t *testing.T) {
 		{"bool", document.NewBoolValue(true), "true"},
 		{"int", document.NewIntegerValue(10), "10"},
 		{"double", document.NewDoubleValue(10.1), "10.1"},
+		{"double with trailing .0", document.NewDoubleValue(10), "10.0"},
```
asdine, line 24: Could you add a test with a very big float with a trailing .0?

yaziine: Done, I removed this one and replaced it with a big float that overflows an int64.

tie: > removed this one

I think we can keep the simple test case.

document/value_test.go
```diff
@@ -21,6 +21,7 @@ func TestValueString(t *testing.T) {
 		{"bool", document.NewBoolValue(true), "true"},
 		{"int", document.NewIntegerValue(10), "10"},
 		{"double", document.NewDoubleValue(10.1), "10.1"},
+		{"double with trailing .0", document.NewDoubleValue(9223372036854775808), "9223372036854775808.0"},
```
tie, line 24: Can we also add a test case for `document.NewDoubleValue(1e24)`? Ideally, this should be `1e+24` (note that JSON integers may not have an exponent). See https://play.golang.org/p/AG47LJ-WEE_0

tie: In particular, this would cover `abs ≥ 1e21` branch where `fmt = 'e'`.

tie: Currently FormatFloat returns `1.0e+24` (that’s still a valid JSON number though).
See https://play.golang.org/p/0qpbJPaeoG_G

tie: Actually, `1e21` would be even better since it’s the boundary for exponent format.

yaziine: Yep good call. I think that it makes sense to have the trailing `.0` even in the `e` notation.

tie (approved):

asdine (approved):

PR #285, opened 2020-10-23
Rename path to reference

asdine:
This PR renames the `ValuePath` type to `Reference` to match the docs


tie:
>A field reference is any sequence of characters that one or more identifiers separated by dots or square brackets.
https://genji.dev/docs/genji-sql/lexical-structure/#field-references

I’m not sure that reference is a good name here. E.g. JSON*Path*, X*Path*. Also, in JSON, [References][1] are not document element paths. Perhaps [Pointer][2] would be a better name?

>[JSON References][1] allows a JSON value to reference another value in a JSON document.
>[JSON Pointer][2] defines a string syntax for identifying a specific value within a JavaScript Object Notation (JSON) document.

[1]: https://tools.ietf.org/id/draft-pbryan-zyp-json-ref-03.html
[2]: https://tools.ietf.org/html/rfc6901

tie:
That said, I don’t think that either makes any difference. Personally I’m more in favor of “path to document element” as it clearly describes the meaning (at least in my opinion). But, again, I think we can use any other term here (pointer, reference, etc) as long as the codebase, docs and everything else is consistent.

asdine:
That's a good point, now that I slept on it I also don't think this is a good idea.
I was bothered by `ValuePath` which sounded weird but using simply "path" sounds great.

I'll close this PR and reopen another one with "path".

PR #286, opened 2020-10-23
Move TestNewFromMap into create_test.go

tzzed:
This PR move some tests  from `document_test.go` into `create_test.go`.

### DIFF
--- a/document/create_test.go
+++ b/document/create_test.go
@@ -73,6 +73,63 @@ func TestNewFromJSON(t *testing.T) {
 	})
 }
 
+func TestNewFromMap(t *testing.T) {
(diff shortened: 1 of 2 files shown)


document/create_test.go
```diff
@@ -73,6 +73,63 @@ func TestNewFromJSON(t *testing.T) {
 	})
 }
 
+func TestNewFromMap(t *testing.T) {
```
asdine, line 76: It must be removed from the other file, it is currently duplicated

asdine:
@tzzed Please change your PR description because if we merge this it will close #260

asdine (approved):
👍🏼

PR #294, opened 2020-10-25
Normalize naming of path related types and functions

asdine:
Most important changes are:
- `document.ValuePath` renamed to `document.Path`
- `expr.FieldSelector` renamed to `expr.Path`

The rest is essentially renaming constructors, tests, etc.


tie (approved):
Neat, that’s simple and concise!

PR #296, opened 2020-10-26
Add support for context.Context APIs

tie:
This PR implements API changes for RFC001.

- Context
  - Remove all previous context.Context usage except for (*query.Query).Exec and sql/driver.
  - genji.New, (*database.DB).BeginTx and (engine.Engine).Begin now accept context.Context.
  - Use WithContext pattern in higher-level API.
  - (*database.DB).Begin uses context.Background() so that we don’t have to update a kazillion of transaction test cases.
- Misc
  - (engine.Engine).Begin now accepts a new engine.TxOptions struct.
  - Rename (engine.Store).NewIterator ⇒ (engine.Store).Iterator (see [Effective Go: Getters](https://golang.org/doc/effective_go.html#Getters))
    Rename engine.IteratorConfig ⇒ engine.IteratorOptions for consistency with other option structs.
  - Add new it.Err() checks wherever Go reports compile error from the previous change.

Note that this PR does not add any new cancellation behavior.

### DIFF
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -196,21 +196,20 @@ func (tx *Transaction) DropTable(name string) error {
 		return errors.New("cannot write to read-only table")
 	}
 
-	it := tx.indexStore.st.NewIterator(engine.IteratorConfig{})
+	it := tx.indexStore.st.Iterator(engine.IteratorOptions{})
+	defer it.Close()
 
 	var buf []byte
 	for it.Seek(nil); it.Valid(); it.Next() {
 		item := it.Item()
-		var opts IndexConfig
 		buf, err = item.ValueCopy(buf)
 		if err != nil {
-			it.Close()
 			return err
 		}
 
+		var opts IndexConfig
 		err = opts.ScanDocument(tx.db.Codec.NewDocument(buf))
 		if err != nil {
-			it.Close()
 			return err
 		}
 
@@ -221,12 +220,10 @@ func (tx *Transaction) DropTable(name string) error {
 
 		err = tx.DropIndex(opts.IndexName)
 		if err != nil {
-			it.Close()
 			return err
 		}
 	}
-	err = it.Close()
-	if err != nil {
+	if err := it.Err(); err != nil {
 		return err
 	}
 
@@ -340,19 +337,19 @@ func (tx *Transaction) ReIndex(indexName string) error {
 
 // ReIndexAll truncates and recreates all indexes of the database from scratch.
 func (tx *Transaction) ReIndexAll() error {
-	var indexes []string
+	it := tx.indexStore.st.Iterator(engine.IteratorOptions{})
+	defer it.Close()
 
-	it := tx.indexStore.st.NewIterator(engine.IteratorConfig{})
+	var indexes []string
 	for it.Seek(nil); it.Valid(); it.Next() {
 		indexes = append(indexes, string(it.Item().Key()))
 	}
-	err := it.Close()
-	if err != nil {
+	if err := it.Err(); err != nil {
 		return err
 	}
 
 	for _, indexName := range indexes {
-		err = tx.ReIndex(indexName)
+		err := tx.ReIndex(indexName)
 		if err != nil {
 			return err
 		}
--- a/database/transaction.go
+++ b/database/transaction.go
@@ -338,7 +338,12 @@ func (tx *Transaction) ReIndex(indexName string) error {
 // ReIndexAll truncates and recreates all indexes of the database from scratch.
 func (tx *Transaction) ReIndexAll() error {
 	it := tx.indexStore.st.Iterator(engine.IteratorOptions{})
-	defer it.Close()
+	itClose := it.Close
+	defer func() {
+		if itClose != nil {
+			itClose()
+		}
+	}()
 
 	var indexes []string
 	for it.Seek(nil); it.Valid(); it.Next() {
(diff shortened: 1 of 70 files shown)


engine/badgerengine/example_test.go
```diff
@@ -12,6 +13,8 @@ import (
 )
 
 func Example() {
+	ctx := context.Background()
+
```
tie, line 17:
```suggestion
```

engine/badgerengine/example_test.go
```diff
@@ -23,7 +26,7 @@ func Example() {
 		log.Fatal(err)
 	}
 
-	db, err := genji.New(ng)
+	db, err := genji.New(ctx, ng)
```
tie, line 29:
```suggestion
	db, err := genji.New(context.Background(), ng)
```

engine/badgerengine/store.go
```diff
@@ -129,19 +129,19 @@ func (s *Store) NextSequence() (uint64, error) {
 
 // NewIterator uses a Badger iterator with default options.
 // Only one iterator is allowed per read-write transaction.
```
tie, line 131:
```suggestion
// Iterator uses a Badger iterator with default options.
// Only one iterator is allowed per read-write transaction.
```

engine/boltengine/store.go
```diff
@@ -72,10 +72,10 @@ func (s *Store) NextSequence() (uint64, error) {
 }
 
 // NewIterator uses the bucket cursor.
```
tie, line 74:
```suggestion
// Iterator uses the Bolt bucket cursor.
```

engine/memoryengine/store.go
```diff
@@ -171,11 +171,11 @@ func (s *storeTx) NextSequence() (uint64, error) {
 	return s.tx.ng.sequences[s.name], nil
 }
 
-func (s *storeTx) NewIterator(cfg engine.IteratorConfig) engine.Iterator {
+func (s *storeTx) Iterator(opts engine.IteratorOptions) engine.Iterator {
```
tie, line 174:
```suggestion
// Iterator creates an iterator with the given options.
func (s *storeTx) Iterator(opts engine.IteratorOptions) engine.Iterator {
```

index/index.go
```diff
@@ -296,6 +296,10 @@ func (idx *Index) iterate(st engine.Store, pivot document.Value, reverse bool, f
 		if err != nil {
 			return err
 		}
+
```
tie, line 299:
```suggestion
```

database/transaction.go
```diff
@@ -340,19 +337,19 @@ func (tx *Transaction) ReIndex(indexName string) error {
 
 // ReIndexAll truncates and recreates all indexes of the database from scratch.
 func (tx *Transaction) ReIndexAll() error {
-	var indexes []string
+	it := tx.indexStore.st.Iterator(engine.IteratorOptions{})
+	defer it.Close()
```
asdine, line 341: This will panic with Badger because only one iterator is allowed at a time. That's why I didn't use a defer here, sorry I should have added a comment.

new.go
```diff
@@ -3,19 +3,22 @@
...
+func New(ctx context.Context, ng engine.Engine) (*DB, error) {
+	db, err := database.New(ctx, ng, database.Options{Codec: msgpack.NewCodec()})
 	if err != nil {
 		return nil, err
 	}
 
 	return &DB{
-		DB: db,
+		DB:      db,
+		Context: ctx,
```
asdine, line 22: We should not pass that context here as it's supposed to be used for controlling the opening of the database.
See https://github.com/genjidb/genji/blob/master/rfcs/001_context.md#open-and-new

asdine (commented):
Great job!

database/transaction.go
tie: Ouch, thanks, you are right.

```suggestion
	itClose := it.Close
	defer itClose()
```

database/transaction.go
```diff
@@ -340,19 +337,19 @@ func (tx *Transaction) ReIndex(indexName string) error {
...
-	it := tx.indexStore.st.NewIterator(engine.IteratorConfig{})
+	var indexes []string
 	for it.Seek(nil); it.Valid(); it.Next() {
 		indexes = append(indexes, string(it.Item().Key()))
 	}
-	err := it.Close()
-	if err != nil {
+	if err := it.Err(); err != nil {
 		return err
 	}
```
tie, line 350:
```suggestion
	// Note that we can’t use deferred close because ReIndex needs
	// an iterator but only one is allowed at a time.
	itClose = func() error { return nil }
	it.Close()

```

new.go
tie: Good call!
```suggestion
		Context: context.Background(),
```

new_wasm.go
```diff
@@ -3,19 +3,22 @@
...
+func New(ctx context.Context, ng engine.Engine) (*DB, error) {
+	db, err := database.New(ctx, ng, database.Options{Codec: custom.NewCodec()})
 	if err != nil {
 		return nil, err
 	}
 
 	return &DB{
-		DB: db,
+		DB:      db,
+		Context: ctx,
```
tie, line 22:
```suggestion
		Context: context.Background(),
```

database/transaction.go
```diff
@@ -340,19 +337,28 @@ func (tx *Transaction) ReIndex(indexName string) error {
...
+	// Note that we can’t use deferred close because ReIndex needs
+	// an iterator but only one is allowed at a time.
+	itClose = nil
+	it.Close()
```
tie, line 358: Actually, this should be in a helper function so that we can defer it.Close() without this weird hack.

asdine (approved):
Thanks! 👏🏼

PR #297, opened 2020-10-27
Initial support of joins(only INNER JOIN)

tdakkota:
Updates #3.


asdine:
Thanks @tdakkota !

We recently introduced the [CONTRIBUTING.md](https://github.com/genjidb/genji/blob/master/.github/CONTRIBUTING.md) file which will help structure a little bit how we are taking contributions.

We want SQL features, especially very important ones like JOINs, to be discussed and planned before accepting any contribution. The purpose is to be certain that we all want the same thing regarding the implementation and that we have thought about most use cases, but also to avoid frustration for contributors who would have spent lots of time working on something that doesn't match the direction of the project.

If you are interested, I invite you to read the [CONTRIBUTING.md](https://github.com/genjidb/genji/blob/master/.github/CONTRIBUTING.md) file and to write a proposal in #3 do that we can start a discussion, does that sound good?

asdine:
Closing this in favor of the new stream API 👍🏼

PR #304, opened 2020-10-30
Fix deadlock with concurrent transactions

### LINKED ISSUE #298: Database is deadlocking
Hi, I've been experimenting with Genji v0.8.0 in one of my apps:

https://github.com/simpleiot/simpleiot/blob/feature-genji2/db/genji/genji.go

After I click around in the frontend a bit, API calls start timing out.

I [instrumented the db calls](https://github.com/simpleiot/simpleiot/commit/3ca34e170d0456388ce180be55e26fc66857696e), and learned that if one of the calls starts before the previous one finishes, the API calls start timing out soon after that.

Am I doing anything obviously wrong? I'm using boltdb backend and thought boltdb was thread safe. Do I need to wrap all db operations in transactions?

asdine:
This fixes #298 by making sure access to the `attachedTxMu` mutex is only locked by Commit or Rollback if that transaction is actually attached.

### DIFF
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -1,0 +1,69 @@
+package database_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/genjidb/genji"
+	"github.com/stretchr/testify/require"
+)
+
+// See issue https://github.com/genjidb/genji/issues/298
+func TestConcurrentTransactionManagement(t *testing.T) {
+	db, err := genji.Open(":memory:")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
+	ch := make(chan struct{})
+	done := make(chan struct{})
+
+	go func() {
+		// 1. Start transaction T1.
+		tx, err := db.Begin(true)
+		require.NoError(t, err)
+
+		// Start transaction T2.
+		ch <- struct{}{}
+		// Wait in case goroutine gets rescheduled.
+		time.Sleep(time.Millisecond)
+
+		// 3. Commit or rollback T1.
+		tx.Rollback()
+
+		// Wait for T2 to finish and return.
+		<-ch
+		done <- struct{}{}
+	}()
+
+	go func() {
+		<-ch // wait for T1 to start.
+
+		// 2. Attempt to start transaction T2.
+		// Waits for T1 to finish.
+		tx, err := db.Begin(true)
+		require.NoError(t, err)
+		defer tx.Rollback()
+
+		ch <- struct{}{}
+	}()
+
+	r := make(chan bool)
+	go func() {
+		t := time.NewTimer(time.Second)
+		select {
+		case <-t.C:
+			r <- false
+		case <-done:
+			r <- true
+		}
+		if !t.Stop() {
+			<-t.C
+		}
+	}()
+
+	if ok := <-r; !ok {
+		t.Fatal("deadlock")
+	}
+}
--- a/database/database_test.go
+++ b/database/database_test.go
@@ -12,7 +12,9 @@ import (
 func TestConcurrentTransactionManagement(t *testing.T) {
 	db, err := genji.Open(":memory:")
 	require.NoError(t, err)
-	defer db.Close()
+	defer func() {
+		require.NoError(t, db.Close())
+	}()
 
 	ch := make(chan struct{})
 	done := make(chan struct{})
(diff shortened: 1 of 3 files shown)


database/database_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}()
+
+	go func() {
+		<-ch // wait for T1 to start.
+
+		// 2. Attempt to start transaction T2.
+		// Waits for T1 to finish.
+		tx, err := db.Begin(true)
+		require.NoError(t, err)
+		defer tx.Rollback()
```
tie, line 45:
```suggestion
		require.NoError(t, tx.Rollback())
```

database/database_test.go
```diff
@@ -0,0 +1,67 @@
...
+		tx, err := db.Begin(true)
+		require.NoError(t, err)
+
+		// Start transaction T2.
+		ch <- struct{}{}
+		// Wait in case goroutine gets rescheduled.
+		time.Sleep(time.Millisecond)
+
+		// 3. Commit or rollback T1.
+		tx.Rollback()
```
tie, line 31:
```suggestion
		require.NoError(t, tx.Rollback())
```

database/database_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/genjidb/genji"
+	"github.com/stretchr/testify/require"
+)
+
+// See issue https://github.com/genjidb/genji/issues/298
+func TestConcurrentTransactionManagement(t *testing.T) {
+	db, err := genji.Open(":memory:")
+	require.NoError(t, err)
+	defer db.Close()
```
tie, line 15:
```suggestion
	defer func() {
		require.NoError(t, db.Close())
	}()
```

tie (commented):
👍

Also, on lines 118–123 in database/database.go, do we need a lock if opts.Attached is false?

```diff
+ if opts.Attached {
	db.attachedTxMu.Lock()
	defer db.attachedTxMu.Unlock()

	if db.attachedTransaction != nil {
		return nil, errors.New("cannot open a transaction within a transaction")
	}
+ }
```

asdine:
> Also, on lines 118–123 in database/database.go, do we need a lock if opts.Attached is false?

@tie Yes because we don't want to allow to any new transaction opening when there is an attached transaction. Existing transaction may continue, but now new ones. https://github.com/genjidb/genji/blob/ac9e33be665939fcfb7c6ad02629916f0ea0faef/database/database.go#L110-L112

tie:
>We don't want to allow to any new transaction opening when there is an attached transaction.

@asdine are there any technical or practical reasons for this limitation? I don’t see how non-attached concurrent transaction `tx` started with `BeginTx` could affect attached one started with `db.Exec("BEGIN")`, i.e. `tx.Exec("BEGIN")` should return “cannot open a transaction within a transaction” error even if there are not transactions attached to database, right?

asdine:
@tie When I started implementing this I followed SQLite behavior:

```go
package main

import (
	"context"
	"database/sql"
	"log"

	_ "github.com/mattn/go-sqlite3"
)

func main() {
	db, err := sql.Open("sqlite3", "./foo.db")
	if err != nil {
		log.Fatal(err)
	}
	defer db.Close()

	_, err = db.Exec("CREATE TABLE IF NOT EXISTS foo(a integer)")
	if err != nil {
		panic(err)
	}

	// start t1 before BEGIN
	tx1, err := db.BeginTx(context.Background(), &sql.TxOptions{
		ReadOnly: true,
	})
	if err != nil {
		panic(err)
	}

	// start BEGIN
	_, err = db.Exec("BEGIN")
	if err != nil {
		panic(err)
	}

	// start t2 after BEGIN
	tx2, err := db.BeginTx(context.Background(), &sql.TxOptions{
		ReadOnly: true,
	})
	if err != nil {
		// panics here with: cannot start a transaction within a transaction
		panic(err)
	}

	err = tx1.Rollback()
	if err != nil {
		panic(err)
	}

	err = tx2.Rollback()
	if err != nil {
		panic(err)
	}

	_, err = db.Exec("ROLLBACK")
	if err != nil {
		panic(err)
	}
}
```

I think we don't have to follow that behavior if we are certain there won't be any weird edge case. I'd be happy to talk about a better design for this, but let's keep the discussion in an issue as this PR is focused on fixing the bug on the v0.8.

tie (approved):

tie:
We’d also need `release-branch.v0.8` branch to cherry-pick this change if we want to release a patch.

PR #306, opened 2020-10-30
Merge key and document packages

asdine:
This PR removes the `key` package and moves its content to 3 different locations:

- Add `MarshalBinary`, `UnmarshalBinary` and `Append` to Value type. These methods encode values without type information.
- Add `ValueEncoder` type which can different types in the same namespace while keeping ordering between types. It is the type used for untyped indexes.
- Add a `pkg/nsb` package for naturally sorted binary representation of supported variables.

This was done because we are going to need the ValueEncoder from within the document package, which currently provokes a cyclic dependency with the key package.


tie:
I don’t think we need any packages under `pkg`. This directory layout pattern is common in Go community because that’s what Go had before *go1.4* release (see https://golang.org/s/go14nopkg). Personally, I don’t consider `pkg` to be idiomatic (unless all packages are under `pkg`, in which case it is redundant).

tie:
What do you think about simple `binarysort` at the root? That would also make the package purpose clearer from the directory listing.

asdine:
I was following the pattern of pushing generic packages under the `pkg` folder before they could eventually get extracted to their own repos (I can't remember where I read this, but it seemed pretty common in Go projects).
However I don't have a strong opinion about this so I don't mind removing it if that seems not idiomatic.

> What do you think about simple binarysort at the root? That would also make the package purpose clearer from the directory listing.

Good idea

binarysort/binarysort.go
```diff
@@ -0,0 +1,137 @@
...
+// so that encoded integers and floats are naturally ordered.
+// Integers will fist be encoded using AppendInt64 on 8 bytes, then 8 zero-bytes will be
+// appended to them.
+// Floats will first be converted to integer, encoded using AppendInt64,
+// then AppendFloat64 will be called with the float value.
+func AppendFloatNumber(buf []byte, x float64) ([]byte, error) {
+	if x > math.MaxInt64 {
+		return AppendFloat64(AppendInt64(buf, math.MaxInt64), x), nil
+	}
+	return AppendFloat64(AppendInt64(buf, int64(x)), x), nil
```
tie, line 136: Shouldn’t we slice the last 8 bytes here?

tie (commented):
Looking through the changes, `naturalsort` would be even better name since the package is more about natural sort order rather than its binary representation. Though `binarysort` is good too.

binarysort/binarysort.go
```diff
@@ -0,0 +1,137 @@
...
+// then AppendFloat64 will be called with the float value.
+func AppendIntNumber(buf []byte, x int64) ([]byte, error) {
+	// appending 8 zero bytes so that the integer has the same size as the double
+	// but always lower for the same value.
+	return append(AppendInt64(buf, x), 0, 0, 0, 0, 0, 0, 0, 0), nil
+}
+
+// AppendFloatNumber takes a number, either integer or float, and encodes it in 16 bytes
+// so that encoded integers and floats are naturally ordered.
+// Integers will fist be encoded using AppendInt64 on 8 bytes, then 8 zero-bytes will be
```
tie, line 128:
```suggestion
// Integers will first be encoded using AppendInt64 on 8 bytes, then 8 zero-bytes will be
```

tie:
It’s a bit unclear from `AppendIntNumber` and `AppendFloatNumber` docs that “number” is encoded as 16 bytes where `float64` is represented as (int64, float64) and `int64` as (int64, zero-padding).

document/value_encoding.go
```diff
@@ -222,7 +221,11 @@ func decodeValueUntil(data []byte, delim, end byte) (Value, int, error) {
 	case BoolValue:
 		i++
 	case DoubleValue:
-		i += 16
+		if data[i+8] == delim {
```
tie, line 224: Are we guaranteed to have `i+8 < len(data) || i+16 < len(data)` here?

tie:
The document decoder is a potential candidate for fuzz tests (should be fine without initial corpus).

binarysort/binarysort.go
```diff
@@ -0,0 +1,126 @@
...
+func AppendBool(buf []byte, x bool) []byte {
+	if x {
+		return append(buf, 1)
+	}
+	return append(buf, 0)
+}
+
+// DecodeBool takes a byte slice and decodes it into a boolean.
+func DecodeBool(buf []byte) bool {
+	return buf[0] == 1
```
tie, line 30: We need a length check here like in `DecodeUint64`. Or document that `DecodeBool` expects non-empty `buf`.

binarysort/binarysort.go
```diff
@@ -0,0 +1,126 @@
...
+		fb ^= 1 << 63
+	} else {
+		fb ^= 1<<64 - 1
+	}
+	return AppendUint64(buf, fb)
+}
+
+// DecodeFloat64 takes a byte slice and decodes it into an float64.
+func DecodeFloat64(buf []byte) (float64, error) {
+	x := binary.BigEndian.Uint64(buf)
```
tie, line 77: `binary.BigEndian.Uint64` will panic if `len(buf) < 8`.

https://github.com/golang/go/blob/846dce9d05f19a1f53465e62a304dea21b99f910/src/encoding/binary/binary.go#L124-L128

binarysort/binarysort.go
```diff
@@ -0,0 +1,126 @@
...
+	b := bytes.NewBuffer(buf)
+	enc := base64.NewEncoder(base64Encoding, b)
+	_, err := enc.Write(data)
+	if err != nil {
+		return nil, err
+	}
+	err = enc.Close()
+	if err != nil {
+		return nil, err
+	}
```
tie, line 99: Am I missing a case with non-nil errors here? It looks like encoder gets error from io.Writer and (*bytes.Buffer).Write always returns nil error.

https://github.com/golang/go/blob/f14119b561cfeefb93e8d773033caeea572dbe71/src/encoding/base64/base64.go#L192-L238
https://github.com/golang/go/blob/f14119b561cfeefb93e8d773033caeea572dbe71/src/bytes/buffer.go#L168-L175

asdine:
Thanks for the review @tie ! I'm closing this PR and opened a new one (https://github.com/genjidb/genji/pull/312) with a different approach that I have been experimenting with these past few weeks. All of your comments have been taken into account in this new branch.

PR #313, opened 2020-11-10
Add cancellation in engines

### LINKED ISSUE #224: Add support for context.Context in engines

asdine:
This PR fixes #224 by adding support for cancellation to engines.

### DIFF
--- a/db.go
+++ b/db.go
@@ -13,14 +13,14 @@ import (
 type DB struct {
 	DB *database.Database
 
-	context context.Context
+	ctx context.Context
 }
 
 // WithContext creates a new database handle using the given context for every operation.
 func (db *DB) WithContext(ctx context.Context) *DB {
 	return &DB{
 		DB:      db.DB,
-		context: ctx,
+		ctx: ctx,
 	}
 }
 
@@ -32,7 +32,7 @@ func (db *DB) Close() error {
 // Begin starts a new transaction.
 // The returned transaction must be closed either by calling Rollback or Commit.
 func (db *DB) Begin(writable bool) (*Tx, error) {
-	tx, err := db.DB.BeginTx(db.context, &database.TxOptions{
+	tx, err := db.DB.BeginTx(db.ctx, &database.TxOptions{
 		ReadOnly: !writable,
 	})
 	if err != nil {
@@ -89,7 +89,7 @@ func (db *DB) Query(q string, args ...interface{}) (*query.Result, error) {
 		return nil, err
 	}
 
-	return pq.Run(db.context, db.DB, argsToParams(args))
+	return pq.Run(db.ctx, db.DB, argsToParams(args))
 }
 
 // QueryDocument runs the query and returns the first document.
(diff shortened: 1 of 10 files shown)


engine/enginetest/testing.go
```diff
@@ -31,7 +31,7 @@ func TestSuite(t *testing.T, builder Builder) {
 	}{
 		{"Engine", TestEngine},
 		{"Transaction/Commit-Rollback", TestTransactionCommitRollback},
-		{"Transaction/Store", TestTransactionStore},
+		{"Transaction/Store", TestTransactionGetStore},
```
tie, line 34:
```suggestion
		{"Transaction/GetStore", TestTransactionGetStore},
```

db.go
```diff
@@ -13,13 +13,14 @@ import (
 type DB struct {
 	DB *database.Database
 
-	Context context.Context
+	context context.Context
```
tie, line 16:
```suggestion
	ctx context.Context
```

db.go
```diff
@@ -13,13 +13,14 @@ import (
...
 }
 
+// WithContext creates a new database handle using the given context for every operation.
 func (db *DB) WithContext(ctx context.Context) *DB {
 	return &DB{
 		DB:      db.DB,
-		Context: ctx,
+		context: ctx,
```
tie, line 23:
```suggestion
		ctx: ctx,
```

db.go
```diff
@@ -31,7 +32,7 @@ func (db *DB) Close() error {
 // Begin starts a new transaction.
 // The returned transaction must be closed either by calling Rollback or Commit.
 func (db *DB) Begin(writable bool) (*Tx, error) {
-	tx, err := db.DB.BeginTx(db.Context, &database.TxOptions{
+	tx, err := db.DB.BeginTx(db.context, &database.TxOptions{
```
tie, line 35:
```suggestion
	tx, err := db.DB.BeginTx(db.ctx, &database.TxOptions{
```

db.go
```diff
@@ -88,7 +89,7 @@ func (db *DB) Query(q string, args ...interface{}) (*query.Result, error) {
 		return nil, err
 	}
 
-	return pq.Run(db.Context, db.DB, argsToParams(args))
+	return pq.Run(db.context, db.DB, argsToParams(args))
```
tie, line 92:
```suggestion
	return pq.Run(db.ctx, db.DB, argsToParams(args))
```

new.go
```diff
@@ -19,6 +19,6 @@ func New(ctx context.Context, ng engine.Engine) (*DB, error) {
 
 	return &DB{
 		DB:      db,
-		Context: context.Background(),
+		context: context.Background(),
```
tie, line 22:
```suggestion
		ctx: context.Background(),
```

tie (approved):
Looks good! Also, we need to update `new_wasm.go` to unexport context; `ctx` is more consistent with embedded contexts in other packages.

PR #322, opened 2020-11-14
Remove in-memory tableInfoStore

asdine:
This PR simplifies how we are handling table information. Currently, table information is loaded in memory on database startup. This was done to fix issues with concurrent table creation with the badger engine, but I believe this is not the right solution for that problem.
With this change, table information is not cached anymore and directly fetched from the engine.

### DIFF
--- a/database/config.go
+++ b/database/config.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-	"sync"
 
 	"github.com/genjidb/genji/document"
 	"github.com/genjidb/genji/engine"
@@ -172,9 +171,6 @@ type TableInfo struct {
 	// name of the store associated with the table.
 	storeName []byte
 	readOnly  bool
-	// if non-zero, this tableInfo has been created during the current transaction.
-	// it will be removed if the transaction is rolled back or set to false if its commited.
-	transactionID int64
 
 	FieldConstraints FieldConstraints
 }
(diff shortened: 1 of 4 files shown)


database/transaction_test.go
```diff
@@ -194,7 +194,7 @@ func TestTxTable(t *testing.T) {
 
 		// Renaming a non existing table should return an error
 		err = tx.AddField("bar", fieldToAdd)
-		require.EqualError(t, database.ErrTableNotFound, err.Error())
+		require.True(t, errors.Is(err, database.ErrTableNotFound))
```
tie, line 197: For consistency with other tests that expect `database.ErrTableNotFound`. This also gives us a better error messages since it prints actual error instead of something like “expected true got false”.

```suggestion
		if !errors.Is(err, database.ErrTableNotFound) {
			require.Equal(t, err, database.ErrTableNotFound)
		}
```

database/config.go
```diff
@@ -316,210 +287,98 @@ func (t *tableInfoStore) Insert(tx *Transaction, tableName string, info *TableIn
...
-
 	return nil
 }
 
-func (t *tableInfoStore) loadAllTableInfo(tx engine.Transaction) error {
-	t.mu.Lock()
-	defer t.mu.Unlock()
-
-	st, err := tx.GetStore([]byte(tableInfoStoreName))
+// Replace table information by the new one.
```
tie, line 363:
```suggestion
// Replace replaces tableName table information with the new info.
```

tie (approved):
Neat!

PR #325, opened 2020-11-15
Move index fetching to bind stage

### LINKED ISSUE #292: sql/planner: tableInputNode should get indexes on Bind for UseIndexBasedOnSelectionNodeRule rule
Optimizer rules should not perform I/O bound operations.

https://github.com/genjidb/genji/blob/a8b144be3e41a21b8d9c5853bfcff5a87a4bb4d7/sql/planner/optimizer.go#L278-L282

asdine:
Fixes #292

### DIFF
--- a/sql/planner/input.go
+++ b/sql/planner/input.go
@@ -15,6 +15,7 @@ type tableInputNode struct {
 
 	tableName string
 	table     *database.Table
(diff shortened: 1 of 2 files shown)


tie (approved):
👍

PR #327, opened 2020-11-16
remove duplicate Reset() of bytes.Buffer

tzzed:
This PR removes duplicates `buf.Reset()`.
`byte.Buffer` is already flushed in `buf.WriteTo(w)`.
  [see WriteTo](https://golang.org/src/bytes/buffer.go?s=7711:7769#L255)


asdine (approved):
Thanks @tzzed !

PR #330, opened 2020-11-19
Fix creating document from struct with known types

Darkclainer:
I think that converting struct to Document is a very complex topic (just look at json/encoding), but #260 introduced behavior that is incompatible with one that was previously (see tests).


asdine (approved):
Thanks @Darkclainer !

PR #332, opened 2020-11-19
Dump command.

### LINKED ISSUE #315: Add dump command
```bash
$ genji dump -h
NAME:
   genji dump - Dump a database or a table as a text file.

USAGE:
   genji dump [options] dbpath

DESCRIPTION:
  The dump command can dump a database as a text file.

  By default, the content of the database is sent to the standard output:

  $ genji dump my.db
  CREATE TABLE foo;
  ...

  It is possible to specify a list of tables to output:

  $ genji dump -t foo -f bar my.db

  The dump command can also write directly into a file:

  $ genji dump -f dump.sql my.db

OPTIONS:
   --table value, -t value   name of the table to dump, it must already exist. Defaults to all tables.
   --file value, -f value       name of the file to output to. Defaults to STDOUT
   --help, -h                show help (default: false)
```

tzzed:
This PR fixes #315 .

### DIFF
--- a/cmd/genji/dump.go
+++ b/cmd/genji/dump.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/genjidb/genji/engine/badgerengine"
+	"github.com/genjidb/genji/engine/boltengine"
+
+	"github.com/dgraph-io/badger/v2"
+	"github.com/genjidb/genji/engine"
+
+	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/cmd/genji/shell"
+)
+
+func executeDump(ctx context.Context, f string, tables []string, e string, dbPath string, w io.Writer) error {
+	if dbPath == "" {
+		return errors.New("db path should be specified")
+	}
+
+	if f != "" {
+		file, err := os.Create(f)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+
+		// file as io.writer for the RunDumpCmd function.
+		w = file
+	}
+
+	var (
+		ng  engine.Engine
+		err error
+	)
+
+	switch e {
+	case "bolt":
+		ng, err = boltengine.NewEngine(dbPath, 0660, nil)
+	case "badger":
+		ng, err = badgerengine.NewEngine(badger.DefaultOptions(dbPath).WithLogger(nil))
+	default:
+		return fmt.Errorf(`engine should be "bolt" or "badger", got %q`, e)
+	}
+	if err != nil {
+		return err
+	}
+
+	// dbPath cannot be empty it is checked in the main.
+	db, err := genji.New(ctx, ng)
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	return shell.RunDumpCmd(db.WithContext(ctx), tables, w)
+}
--- a/cmd/genji/dump.go
+++ b/cmd/genji/dump.go
@@ -2,10 +2,8 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
-	"os"
 
 	"github.com/genjidb/genji/engine/badgerengine"
 	"github.com/genjidb/genji/engine/boltengine"
@@ -17,22 +15,7 @@ import (
 	"github.com/genjidb/genji/cmd/genji/shell"
 )
 
-func executeDump(ctx context.Context, f string, tables []string, e string, dbPath string, w io.Writer) error {
-	if dbPath == "" {
-		return errors.New("db path should be specified")
-	}
-
-	if f != "" {
-		file, err := os.Create(f)
-		if err != nil {
-			return err
-		}
-		defer file.Close()
-
-		// file as io.writer for the RunDumpCmd function.
-		w = file
-	}
-
+func executeDump(ctx context.Context, w io.Writer, tables []string, e, dbPath string) error {
 	var (
 		ng  engine.Engine
 		err error
@@ -50,12 +33,11 @@ func executeDump(ctx context.Context, f string, tables []string, e string, dbPat
 		return err
 	}
 
-	// dbPath cannot be empty it is checked in the main.
 	db, err := genji.New(ctx, ng)
 	if err != nil {
 		return err
 	}
 	defer db.Close()
 
-	return shell.RunDumpCmd(db.WithContext(ctx), tables, w)
+	return shell.RunDumpCmd(db.WithContext(ctx), w, tables)
 }
--- a/cmd/genji/dump.go
+++ b/cmd/genji/dump.go
@@ -8,7 +8,7 @@ import (
 	"github.com/genjidb/genji/engine/badgerengine"
 	"github.com/genjidb/genji/engine/boltengine"
 
-	"github.com/dgraph-io/badger/v2"
+	"github.com/dgraph-io/badger/v3"
 	"github.com/genjidb/genji/engine"
 
 	"github.com/genjidb/genji"
--- a/cmd/genji/dump_test.go
+++ b/cmd/genji/dump_test.go
@@ -1,0 +1,117 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/cmd/genji/shell"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExecuteDump(t *testing.T) {
+	// db to save.
+	testDb := os.TempDir() + "test.db"
+	file := os.TempDir() + "dump.sql"
+	tests := []struct {
+		name   string
+		tables []string
+		file   string
+		engine string
+		dbPath string
+		fails  bool
+	}{
+		{"Dump errored with no option", []string{}, ``, "bolt", ``, true},
+		{"Dump errored with bad engine", []string{"test", "foo"}, file, "test", testDb, true},
+		{"Dump to stdout", []string{"test"}, ``, "bolt", testDb, false},
+		{"Dump to stdout list of tables", []string{"test", "foo"}, ``, "bolt", testDb, false},
+		{"Dump to stdout list of tables with badger", []string{"test", "foo"}, ``, "badger", os.TempDir(), false},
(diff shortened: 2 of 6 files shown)


cmd/genji/main.go
```diff
@@ -111,6 +112,57 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
...
+					Value:   "bolt",
+				},
+			},
+			Action: func(c *cli.Context) error {
+				table := c.StringSlice("table")
+				file := c.String("file")
+				engine := c.String("engine")
+				args := c.Args().First()
+
+				if err := executeDump(context.Background(), file, table, engine, args, os.Stdout); err != nil {
```
yaziine, line 159: It's better to forward the context received in parameter.

```suggestion
				if err := executeDump(c.Context, file, table, engine, args, os.Stdout); err != nil {
```

cmd/genji/dump.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/genjidb/genji/engine"
+
+	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/cmd/genji/shell"
+)
+
+func executeDump(ctx context.Context, f string, tables []string, e string, dbPath string, w io.Writer) error {
+	if dbPath == "" {
+		return errors.New("db path should be specified")
+	}
```
yaziine, line 23: I think that this can be checked earlier in the execution flow.

cmd/genji/dump.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/genjidb/genji/engine/boltengine"
+
+	"github.com/dgraph-io/badger/v2"
+	"github.com/genjidb/genji/engine"
+
+	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/cmd/genji/shell"
+)
+
+func executeDump(ctx context.Context, f string, tables []string, e string, dbPath string, w io.Writer) error {
```
yaziine, line 20: I don't think we need to pass `os.Stdout` as a parameter here.

```suggestion
func executeDump(ctx context.Context, f string, tables []string, e string, dbPath string) error {
```

cmd/genji/dump.go
```diff
@@ -0,0 +1,61 @@
...
+	case "badger":
+		ng, err = badgerengine.NewEngine(badger.DefaultOptions(dbPath).WithLogger(nil))
+	default:
+		return fmt.Errorf(`engine should be "bolt" or "badger", got %q`, e)
+	}
+	if err != nil {
+		return err
+	}
+
+	// dbPath cannot be empty it is checked in the main.
```
yaziine, line 53: I think this comment can be removed?
```suggestion
```

cmd/genji/main.go
```diff
@@ -111,6 +112,57 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
 				return nil
 			},
 		},
+		{
+			Name:      "dump",
+			Usage:     "Dump a database or a list of tables as a text file.",
+			UsageText: `genji dump [options] dbpath`,
+
```
yaziine, line 119:
```suggestion
```

cmd/genji/main.go
```diff
@@ -111,6 +112,57 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
...
+				},
+			},
+			Action: func(c *cli.Context) error {
+				table := c.StringSlice("table")
+				file := c.String("file")
+				engine := c.String("engine")
+				args := c.Args().First()
+
+				if err := executeDump(context.Background(), file, table, engine, args, os.Stdout); err != nil {
+					cli.ShowCommandHelpAndExit(c, c.Command.Name, 2)
```
yaziine, line 160: If any non-nil error is returned, the help will be output. I am not sure that we want this behaviour. For example do we want to show the help if the engine cannot be instantiated?

cmd/genji/shell/command.go
```diff
@@ -315,22 +314,22 @@ func runDumpCmd(db *genji.DB, tables []string, w io.Writer) error {
...
+			}
+		}
+
+		if err = dumpTable(tx, table, w); err != nil {
 			// If table doesn’t exist we skip it.
 			if errors.Is(err, database.ErrTableNotFound) {
 				continue
 			}
-			_, err = fmt.Fprintln(w, "COMMIT;")
+			_, err = fmt.Fprintln(w, "ROLLBACK;")
```
yaziine, line 329: I am not sure to understand this changes here. Could you elaborate please?

cmd/genji/dump_test.go
```diff
@@ -0,0 +1,117 @@
...
+		{"Dump to stdout list of tables", []string{"test", "foo"}, ``, "bolt", testDb, false},
+		{"Dump to stdout list of tables with badger", []string{"test", "foo"}, ``, "badger", os.TempDir(), false},
+		{"Dump in a file", []string{"test", "foo"}, file, "bolt", testDb, false},
+		{"Dump in a file with badger", []string{"test", "foo"}, file, "badger", os.TempDir(), false},
+	}
+
+	for _, tt := range tests {
+		t.Cleanup(func() {
+			_ = os.RemoveAll(tt.dbPath)
+			_ = os.RemoveAll(testDb)
```
yaziine, line 40: I think removing `tt.dbPath` is enough. Isn't it?

cmd/genji/dump_test.go
```diff
@@ -0,0 +1,117 @@
...
+				b, err := ioutil.ReadFile(tt.file)
+				require.NoError(t, err)
+				buf.Write(b)
+			}
+
+			if tt.fails {
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
```
yaziine, line 108: This block can be moved above, close to the `executeDump` call.

cmd/genji/dump_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+			ci := "COMMIT;\n"
+			b.WriteString(ci)
```
yaziine, line 111:
```suggestion
			b.WriteString("COMMIT;\n")
```

yaziine: Is this can be moved above just after the for loop?

cmd/genji/dump_test.go
```diff
@@ -0,0 +1,117 @@
...
+		})
+
+		t.Run(tt.name, func(t *testing.T) {
+			db, err := genji.Open(":memory:")
+			require.NoError(t, err)
+
+			defer db.Close()
+			var b bytes.Buffer
+			tx := "BEGIN TRANSACTION;\n"
+			b.WriteString(tx)
```
yaziine, line 50:
```suggestion
			b.WriteString("BEGIN TRANSACTION;\n")
```

yaziine (commented):
Thanks for handling this @tzzed! Overall, it looks good 👍 

I've left some comments.

cmd/genji/dump.go
tzzed: At the first I did the check in the main, but after some reflection I thought it is better to check in the here. Let's imagine tomorrow we write a function or command that call `executeDump`?

cmd/genji/dump_test.go
tzzed: Yes good catch.

cmd/genji/dump_test.go
tzzed: As it is the very last thing that is write, is more "safe" to write it here. If we want to add some test we won't care about it.

cmd/genji/shell/command.go
tzzed: If an error occurs during the dump we should `ROLLBACK`, it's a correction.

cmd/genji/dump.go
tzzed: I kept the same logic as the `RunDumpCommand` it is better for testing... And the same argument that above.

yaziine: I suggest to have something like this so that you can easily test
```go
func executeDump(ctx context.Context, w io.Writer, tables []string, e, dbPath string) error
```

cmd/genji/dump.go
yaziine: > should we trust the caller

Since that we have only one entry point and we are the caller I think we can trust it.

> Do you think it 's better to make the check before?

Honestly, I don't have a strong opinion on this, it's up to you.

cmd/genji/dump_test.go
yaziine: ISTM that you're not writing anything after the loop so it makes sense to me that the last write is just after the loop.

cmd/genji/dump_test.go
```diff
@@ -0,0 +1,117 @@
...
+			require.NoError(t, err)
+
+			var buf bytes.Buffer
+			err = executeDump(context.Background(), tt.file, tt.tables, tt.engine, tt.dbPath, &buf)
+
+			if tt.file != "" {
+				b, err := ioutil.ReadFile(tt.file)
+				require.NoError(t, err)
+				buf.Write(b)
+			}
```
yaziine, line 102: I am wondering if we do need this? The content of the file is never checked. What do you think?

cmd/genji/shell/command.go
yaziine: Oops, got it!

engine/badgerengine/go.sum
```diff
@@ -3,15 +3,19 @@ github.com/DataDog/zstd v1.4.1 h1:3oxKN3wbHibqx897utPC2LTQU4J+IHWWJO+glkAkpFM=
 github.com/DataDog/zstd v1.4.1/go.mod h1:1jcaCB/ufaK+sKp1NBhlGmpz41jOoPQ35bpF36t7BBo=
 github.com/OneOfOne/xxhash v1.2.2 h1:KMrpdQIwFcEqXDklaen+P1axHaj9BSKzvpUUfnHldSE=
 github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAElWljhcU=
+github.com/agnivade/levenshtein v1.1.0/go.mod h1:veldBMzWxcCG2ZvUTKD2kJNRdCk5hVbJomOvKkmgYbo=
```
asdine, line 6: The badger engine should not import the cli dependencies. Could you revert all the dependencies changes of this PR?

asdine (commented):
Thanks @tzzed , it looks good 👍🏼 
Could you rebase with master please?

asdine (approved):
Thanks @tzzed !

PR #347, opened 2021-02-10
Lazyness .save command

tzzed:
This PR is a laziness version of `.save` command using `io.writer` redirection.


asdine:
As discussed with @tzzed, this change is not needed. Thanks @tzzed !

Issue #321, opened 2020-11-14
Add restore command 


asdine:
```bash
$ genji restore -h
NAME:
   genji restore - Restore a database or a table from a file created by genji dump

USAGE:
   genji restore dumpFile dbPath

DESCRIPTION:
  The restore command can restore a database or a table from a text file.

  $ genji restore dump.sql my.db 

OPTIONS:
   --help, -h                show help (default: false)
```

yaziine:
I think it should take more flags for specifying the table and the engine to use:
```
$ genji restore -h
NAME:
   genji restore - Restore a database or a table from a file created by genji dump

USAGE:
   genji restore [options] dumpFile dbPath

DESCRIPTION:
  The restore command can restore a database or a table from a text file.

  $ genji restore dump.sql my.db 

OPTIONS:
   --engine, -e            name of the engine to use, options are 'bolt' or 'badger'. Default to 'bolt'
   --table, -t             name of the table to restore. Default to all tables
   --help, -h              show help (default: false)
```

tzzed:
I think the `--table` option is not usefull in this case. Restoring a table from a "dump" file seems weird. In my point of view, it will be more simplest and usefull in a `.restore` command.
What is the expected behavior if the `dbPath` exists? Should the file be overwritten? return an error?

asdine:
> I think the --table option is not usefull in this case

Agreed

> What is the expected behavior if the dbPath exists? Should the file be overwritten? Return an error? It currently returns error: table already exists for the CREATE query.

I think this is good enough, we can leave it like that

PR #357, opened 2021-03-01
Add restore command.

tzzed:
This a proposal for #321 . This PR allows only the full database restoration from a file.

### DIFF
--- a/cmd/genji/main.go
+++ b/cmd/genji/main.go
@@ -182,13 +182,8 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
 			UsageText: `genji restore dumpFile dbPath`,
 			Description: `The restore command can restore a database from a text file.
 
-		$ genji restore -f dump.sql -e bolt my.db`,
+		$ genji restore dump.sql -e bolt my.db`,
 			Flags: []cli.Flag{
-				&cli.StringFlag{
-					Name:    "file",
-					Aliases: []string{"f"},
-					Usage:   "name of the source file restoration.",
-				},
 				&cli.StringFlag{
 					Name:    "engine",
 					Aliases: []string{"e"},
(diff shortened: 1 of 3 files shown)


cmd/genji/main.go
```diff
@@ -176,6 +176,48 @@ $ curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db my.
...
+			Usage:     "Restore a database from a file created by genji dump",
+			UsageText: `genji restore dumpFile dbPath`,
+			Description: `The restore command can restore a database from a text file.
+
+		$ genji restore -f dump.sql -e bolt my.db`,
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:    "file",
+					Aliases: []string{"f"},
+					Usage:   "name of the source file restoration.",
```
asdine, line 190: Why do we need this option? cf https://github.com/genjidb/genji/issues/321#issuecomment-730380223

cmd/genji/restore.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/dgraph-io/badger/v3"
+	"github.com/genjidb/genji"
+	"github.com/genjidb/genji/engine"
+	"github.com/genjidb/genji/engine/badgerengine"
+	"github.com/genjidb/genji/engine/boltengine"
+)
+
+func executeRestore(ctx context.Context, r io.Reader, e, dbPath string) error {
+	if dbPath == "" {
```
asdine, line 18: No need to double check, we already did in the other file

cmd/genji/restore.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return err
+	}
+	db = db.WithContext(ctx)
+	defer db.Close()
+
+	scanner := bufio.NewScanner(r)
+
+	// Every query ends with a semicolon.
+	scanner.Split(func(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
asdine, line 49: Nice 👍🏼 
It makes me think that Genji should provide a way to execute queries from a reader

cmd/genji/restore_test.go
```diff
@@ -0,0 +1,122 @@
...
+		name   string
+		engine string
+		path   string
+		target string
+		fail   bool
+	}{
+		{"without engine", "", "", "", true},
+		{"error with bad engine", "test", "", "", true},
+		{"error without db path", "bolt", "", "", true},
+		{"with bolt", "bolt", "test.db", "restored.db", false},
```
asdine, line 28: We should use a temporary directory for this, using ioutil.TempDir. https://golang.org/pkg/io/ioutil/#TempDir

Here's an example with Bolt: https://github.com/genjidb/genji/blob/main/engine/boltengine/engine_test.go#L37-L42

asdine (approved):
Thanks @tzzed ! I'll merge and reorganize the cli code so we don't need these tests anymore

PR #374, opened 2021-03-17
Add .schema command.

### LINKED ISSUE #339: Add .schema command
### Proposal

This is a proposal of `.schema` command to show the `CREATE` statements matching pattern.
The command shows all tables as default or shows the table name option.

Expected: 
```sql
genji> .schema  [table_name]
CREATE TABLE foo (
    bar text NOT NULL,
   baz text NOT NULL
);

CREATE INDEX idx_foo_bar ON foo (bar);

CREATE TABLE bar (
    baz text NOT NULL,
);


```

### Motivation

Like the `.schema`  command of sqlite this is a good command to see the schema of all tables.

tzzed:
Fixes #339.

### DIFF
--- a/cmd/genji/dbutil/dump.go
+++ b/cmd/genji/dbutil/dump.go
@@ -53,7 +53,7 @@ func Dump(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) erro
 			return err
 		}
 
-		return dumpTable(tx, tableName, w)
+		return dumpTableContent(tx, w, tableName)
 	})
 	if err != nil {
 		_, er := fmt.Fprintln(w, "ROLLBACK;")
@@ -64,10 +64,86 @@ func Dump(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) erro
 	return err
 }
 
-// dumpTable displays the content of the given table as SQL statements.
-func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
+// dumpTableContent displays the content of the given table as SQL statements.
+func dumpTableContent(tx *genji.Tx, w io.Writer, tableName string) error {
+	// Dump schema first.
+	if err := dumpSchema(tx, w, tableName); err != nil {
+		return err
+	}
+
+	t, err := tx.GetTable(tableName)
+	if err != nil {
+		return err
+	}
+
+	q := fmt.Sprintf("SELECT * FROM %s", t.Name())
+	res, err := tx.Query(q)
+	if err != nil {
+		return err
+	}
+	defer res.Close()
+
 	var buf bytes.Buffer
 
+	// Inserts statements.
+	insert := fmt.Sprintf("INSERT INTO %s VALUES ", t.Name())
+	return res.Iterate(func(d document.Document) error {
+		buf.WriteString(insert)
+
+		data, err := document.MarshalJSON(d)
+		if err != nil {
+			return err
+		}
+
+		buf.Write(data)
+		buf.WriteString(";\n")
+		_, err = buf.WriteTo(w)
+		return err
+	})
+}
+
+// DumpSchema takes a database and dumps its schema as SQL queries in the given writer.
+// If tables is provided, only selected tables will be outputted.
+func DumpSchema(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) error {
+	tx, err := db.Begin(false)
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
+
+	query := "SELECT table_name FROM __genji_tables"
+	if len(tables) > 0 {
+		query += " WHERE table_name IN ?"
+	}
+
+	res, err := tx.Query(query, tables)
+	if err != nil {
+		return err
+	}
+	defer res.Close()
+
+	i := 0
+	return res.Iterate(func(d document.Document) error {
+		// Blank separation between tables.
+		if i > 0 {
+			if _, err := fmt.Fprintln(w, ""); err != nil {
+				return err
+			}
+		}
+		i++
+
+		// Get table name.
+		var tableName string
+		if err := document.Scan(d, &tableName); err != nil {
+			return err
+		}
+
+		return dumpSchema(tx, w, tableName)
+	})
+}
+
+// dumpSchema displays the schema of the given table as SQL statements.
+func dumpSchema(tx *genji.Tx, w io.Writer, tableName string) error {
 	t, err := tx.GetTable(tableName)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


cmd/genji/dbutil/dump.go
```diff
@@ -64,10 +64,86 @@ func Dump(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) erro
...
-// dumpTable displays the content of the given table as SQL statements.
-func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
+// dumpTableContent displays the content of the given table as SQL statements.
+func dumpTableContent(tx *genji.Tx, w io.Writer, tableName string) error {
+	// Dump schema first.
+	if err := dumpSchema(tx, w, tableName); err != nil {
+		return err
+	}
+
+	t, err := tx.GetTable(tableName)
```
asdine, line 74: We can avoid getting the table here, since we already have `tableName`. If the table doesn't exist, we'll get an error below anyway

cmd/genji/dbutil/dump.go
```diff
@@ -64,10 +64,86 @@ func Dump(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) erro
...
+		buf.WriteString(insert)
+
+		data, err := document.MarshalJSON(d)
+		if err != nil {
+			return err
+		}
+
+		buf.Write(data)
+		buf.WriteString(";\n")
+		_, err = buf.WriteTo(w)
```
asdine, line 100: Using a bytes.Buffer to then write to w is useless here, instead we can replace it by a bufio.Writer https://golang.org/pkg/bufio/#Writer or fmt.Fprintf (see DumpSchema for an example)

cmd/genji/dbutil/dump.go
```diff
@@ -64,10 +64,86 @@ func Dump(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) erro
...
+		return err
+	})
+}
+
+// DumpSchema takes a database and dumps its schema as SQL queries in the given writer.
+// If tables is provided, only selected tables will be outputted.
```
asdine, line 106:
```suggestion
// If tables are provided, only selected tables will be outputted.
```

cmd/genji/dbutil/dump.go
```diff
@@ -64,10 +64,86 @@ func Dump(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) erro
 	return err
 }
 
-// dumpTable displays the content of the given table as SQL statements.
-func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
+// dumpTableContent displays the content of the given table as SQL statements.
+func dumpTableContent(tx *genji.Tx, w io.Writer, tableName string) error {
+	// Dump schema first.
+	if err := dumpSchema(tx, w, tableName); err != nil {
```
asdine, line 70: The name of the function indicates that it only dumps the table content, not schema. Either we keep the same name as before, or we remove the `dumpSchema` call from this function and call it separately

cmd/genji/dbutil/dump_test.go
```diff
@@ -88,3 +88,65 @@ func TestDump(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			db, err := genji.Open(":memory:")
+			require.NoError(t, err)
+			defer db.Close()
+
+			var want bytes.Buffer
+
+			getBuffer := func(table string) func(s string) {
```
asdine, line 109: This was only necessary in the scope of the original test, it is not needed here

cmd/genji/shell/command.go
```diff
@@ -55,6 +55,12 @@ var commands = []struct {
 		DisplayName: ".save",
 		Description: "Save database content in the specified file.",
 	},
+	{
+		Name:        ".schema",
+		Options:     "[table_name]",
+		DisplayName: ".schema",
+		Description: "Show the CREATE statements matching pattern of a database or the given tables.",
```
asdine, line 62:
```suggestion
		Description: "Show the CREATE statements of all tables or of the selected ones",
```

cmd/genji/dbutil/dump.go
tzzed: Yes good catch !

cmd/genji/dbutil/dump.go
tzzed: You're right , it's better to keep as before `dumpTable` and no need to make a new function `dumpTableContent`.

cmd/genji/dbutil/dump_test.go
tzzed: It is necessary to keep the alphabetic order of the `tt.tables` else we should sort the slice of `tt.tables` manually.

asdine: If we keep it, it would be useful to change l98 to `[]string{"tblA", "tblB", "tblC"}`, otherwise we won't have any output

cmd/genji/dbutil/dump.go
```diff
@@ -53,7 +54,7 @@ func Dump(ctx context.Context, db *genji.DB, w io.Writer, tables ...string) erro
 			return err
 		}
 
-		return dumpTable(tx, tableName, w)
+		return dumpTable(tx, bufio.NewWriter(w), tableName)
```
asdine, line 57: I don't think there is a point into:
- passing around the bufio.Writer as param (l.69)
- using the bufio writer after we already written to it (l19 to 56)

We should:

- create a bufio.Writer l19 and l104
- pass io.Writer everywhere
- use `fmt.Fprint*` everywhere

asdine (approved):
Good job @tzzed !
Could you squash your branch into one or two relevant commits and force push please?

cmd/genji/dbutil/dump.go
```diff
@@ -128,32 +206,13 @@ func dumpTable(tx *genji.Tx, tableName string, w io.Writer) error {
 			u = " UNIQUE"
 		}
 
-		_, err = fmt.Fprintf(w, "CREATE%s INDEX %s ON %s (%s);\n", u, index.Info.IndexName, index.Info.TableName,
+		infos := stringutil.Sprintf("CREATE%s INDEX %s ON %s (%s);", u, index.Info.IndexName, index.Info.TableName,
```
asdine, line 209: We can use `fmt.Fprintf` directly here, `stringutil.Sprintf` is only meant to be used in the core of Genji to reduce the size of compiled Wasm

tzzed: Good catch.

Issue #383, opened 2021-04-11
sql driver: conversion of lazy loaded document when Scanning


paralin:
Now that #382 was merged, my go-orm test is proceeding past inserting objects into the db!

The next hurdle is getting Scan() to work:

```
panic: sql: Scan error on column index 0, name "*": unsupported Scan, storing driver.Value type *database.lazilyDecodedDocument into type *sql.RawBytes
```

In database/sql/convert.go:219 (convertAssignRows) it tries to convert the *lazilyDecodedDocument into *sql.RawBytes.

asdine:
Cool! Could you share the query you tried to run? And any info that could help me investigate

gedw99:
hey @paralin where is the ORM test code.

Are you using this for skiffOS ?

paralin:
@asdine I've posted a PR with the ORM test code in #390

gedw99:
@paralin PR looks pretty nice. Will have a look

gedw99:
@asdine yep

PR #390, opened 2021-04-18
[WIP] go-orm proof of concept example

paralin:
Currently fails due to issue #383 

This is a proof-of-concept of constructing a Go-orm instance backed by GenjiDB.


paralin:
https://github.com/paralin/genji/tree/gorm-example-mysql-dialect

Here's a variant which uses the mysql dialect instead of the sqlite one. I think `AUTO_INCREMENT` is not supported (right?) so that part is disabled.

asdine:
Thanks @paralin ! I would love to make Genji compatible with Gorm! I think what you did is more than just an example and deserves a separate repo. I propose to create a repo in the genjidb org so then you can create a PR there, what do you think?

paralin:
@asdine Sorry never saw your reply for some reason, whatever you want to do with this example / demo, let me know.

paralin:
I think there are enough features to make it work; at least last time I tried it

But I'd have to spend some time digging through the code to figure out how I made it work.

If anyone really needs this urgently feel free to ping here and I'll see if I can post a repro.

asdine:
@paralin There is no rush, I don't yet have time to tackle this issue, if you want to do it let me know at anytime and I'll re-open it and answer any question you may have

PR #403, opened 2021-05-22
Store schemas as SQL

asdine:
This PR changes the way table and index schemas are stored. Previously, we used to store the parsed representation (*database.TableInfo, *database.IndexInfo) as custom documents.
Instead, we now store the table information in SQL form and parse it at startup.
The motivation behind this change is to reduce the areas that could lead to a breaking change in the future. That way, if we decide to improve the internal representation of table and index information, the SQL representation won't change, neither the storage structure.
This also adds support for prepared statements.


cmd/genji/dbutil/dump.go
```diff
@@ -135,91 +134,67 @@ func DumpSchema(ctx context.Context, db *genji.DB, w io.Writer, tables ...string
...
-
-		_, err = fmt.Fprintf(w, "CREATE%s INDEX %s ON %s (%s);\n", u, index.Info.IndexName, index.Info.TableName, strings.Join(paths, ", "))
+		res, err := tx.Query(q, tableName)
 		if err != nil {
 			return err
 		}
-	}
+		defer res.Close()
+
+		err = res.Iterate(func(d document.Document) error {
```
tzzed, line 187: @asdine what do you think about `return res.Iterate(func(d document.Document)` ?

stream/stream.go
```diff
@@ -82,6 +82,18 @@ func (s *Stream) String() string {
 	return sb.String()
 }
 
+func (s *Stream) Clone() *Stream {
+	var clone *Stream
+
+	n := s.First()
+	for n != nil {
```
tzzed, line 89: We can do  one line `for n := s.First(); n != nil ; n = n.GetNext()`

tzzed (commented):
@asdine GG 👍  ! I allowed myself to review this PR and I did some suggestions.

asdine:
Thanks @tzzed !

PR #412, opened 2021-06-12
Fix README example

maxwondercorn:
The example using `database/sql` does not have the correct import path for the driver

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ func main() {
 
 ```go
 // import Genji as a blank import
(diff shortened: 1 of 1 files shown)


asdine:
Thanks @maxwondercorn ! Actually the README is sync'ed with the current state of the main branch, which rearranged the packages. You can find the README for the latest version here https://github.com/genjidb/genji/blob/v0.12.0/README.md.

I know this is not ideal, as new users will always see the README of the main branch, but package layout is unlikely to change much in the future

PR #414, opened 2021-06-20
Use document.Iterator for scanDocument

gillesfabio:
👋 @asdine,

Genji is sweet ❤️ You rock, man.

It seems `scanDocument` only needs to call `Iterate` on `*Result`. Why not passing a `document.Iterator` instead? As the function is called `scanDocument` and returns a `document.Document`, taking a `document.Iterator` as argument looks a bit more `document` API friendly.

Cheers 🏖️


asdine (approved):
Thanks @gillesfabio ! 🙏🏼

PR #415, opened 2021-06-21
Remove unused code from document/FieldBuffer.Delete

gillesfabio:
In `document/FieldBuffer.Delete`, the return value of `parentPath[:len(parentPath)-1].GetValueFromDocument(fb)` is never used. I suspect a forgotten line or an unhandled case that should be handled (and I'm ready to work on this issue if so).


asdine (approved):
Thanks @gillesfabio ! I think this is just a left-over from a refactoring 👍🏼

PR #441, opened 2022-02-06
CloneValue: use copy instead of  iterate.

tzzed:
Use copy instead of iterate.


asdine (approved):
Good catch @tzzed thanks!

PR #442, opened 2022-02-08
return nil instead of err 

tzzed:


asdine (commented):
Could you explain a little bit more in the description what this PR fixes?

It appears that even if that variable is not used, the code is still correct.
As much as possible, I would like to avoid small PRs that don't change or fix the behavior of the project.

tzzed:
Yes it doesn't change the anything the code was correct. In reality, this PR doesn't have a real value.

Issue #443, opened 2022-02-10
Display execution time.


tzzed:
### Proposal
Currently there is no way to know the execution's time of a query.
As Sqlite with a command `.timer on|off`. 

```shell
genji>.timer on
genji>select * from foo;
{
  "Username": "Tonydorrf"
}
Time: 0ms
genji>
```
Then turn off the timer
```shell
genij>.timer off
genji>select * from foo;
{
  "Username": "Tonydorrf"
}
genji>
```

### Motivation
Sometimes users need to measure performances of requests, index efficiency....
Sqlite proposes the command `.timer on|off` which measures the time of every command while the timer is active `.timer on` and it doesn't after turning off `.timer off`.
The timer is off by default.

asdine:
Good idea! Interested?

tzzed:
Yes!

asdine:
Closed by #445

PR #445, opened 2022-02-11
Add .timer command.

tzzed:
This is a proposal for #443.

### DIFF
--- a/cmd/genji/shell/shell.go
+++ b/cmd/genji/shell/shell.go
@@ -200,27 +200,11 @@ func (sh *Shell) runExecutor(ctx context.Context, promptExecCh chan string) erro
 
 			// if showtime is true, ensure it's a query, and it was executed.
 			if showTime && !strings.HasPrefix(input, ".") && strings.HasSuffix(input, ";") {
-				displayTime(start)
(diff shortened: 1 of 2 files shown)


cmd/genji/shell/shell.go
```diff
@@ -189,12 +195,34 @@ func (sh *Shell) runExecutor(ctx context.Context, promptExecCh chan string) erro
...
+			if showTime && !strings.HasPrefix(input, ".") && strings.HasSuffix(input, ";") {
+				displayTime(start)
 			}
 		case promptExecCh <- "":
 		}
 	}
 }
 
+// displayTime shows execution time of a query.
+func displayTime(start time.Time) {
```
asdine, line 211: How about using the default display of the [time.Duration](https://pkg.go.dev/time#Duration.String) type?

asdine (commented):
Thanks @tzzed ! Just one comment

cmd/genji/shell/shell.go
asdine: I think we can just use the time.Duration output yeat

tzzed: Simple and efficient!

asdine (approved):
✨

PR #448, opened 2022-02-12
use transaction in InsertJSON

tzzed:
InsertJson is used to insert json data into a db but if something wrong happens we should rollback to the initial state of the database to avoid partial insertion. 
Especially in the case of the following usage. 
`curl https://api.github.com/repos/genjidb/genji/issues | genji insert --db mydb -t foo`

### DIFF
--- a/cmd/genji/commands/insert.go
+++ b/cmd/genji/commands/insert.go
@@ -10,6 +10,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/urfave/cli/v2"
 
+	"github.com/genjidb/genji"
 	"github.com/genjidb/genji/cmd/genji/dbutil"
 )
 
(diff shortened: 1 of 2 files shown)


asdine:
@tzzed Is there a reason why this was closed? I think this change is a good idea

tzzed:
After reflection, I though to create an issue was a better way. see #449

internal/sql/parser/drop.go
```diff
@@ -39,7 +41,7 @@ func (p *Parser) parseDropTableStatement() (statement.DropTableStmt, error) {
 	// Parse table name
 	stmt.TableName, err = p.parseIdent()
 	if err != nil {
-		pErr := err.(*ParseError)
+		pErr := errors.Unwrap(err).(*ParseError)
```
asdine, line 44: Could you rebase your branch please? This has already been merged

tzzed: done.

asdine (approved):
Good job!

PR #452, opened 2022-02-24
use fieldbuffer copy

tzzed:
little refactoring use `fb.Copy(d)` instead of iterate.

### DIFF
--- a/document/create.go
+++ b/document/create.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/buger/jsonparser"
 	"github.com/cockroachdb/errors"
+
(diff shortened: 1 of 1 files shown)


asdine:
Thank you @tzzed, but as mentioned in #442 I'd like to avoid small refactoring PRs.
The project is still in development phase so small refactoring are not really useful for several reasons:
- This part of the code might move or be removed
- This `types.Document` API, or any other API might change in the near future

I agree that having a clean codebase is important and I understand why you made this change, but it's too early for this project and it would be better if it was grouped with other changes in a bigger cleanup PR.

Issue #454, opened 2022-03-28
badgerengine doesn't release disk space since badger gc doesn't work.


crazycs520:
### What version of Genji are you using?

<pre>
$ genji version
v0.14.1
</pre>

### What did you do?

Related issue: https://discuss.dgraph.io/t/badgerdb-consume-too-much-disk-space/17070/2

This problem is caused when `Transaction` `Commit`,  the `Transaction` doesn't call `tx.Discard()` then cause badger gc doesn't work. Since the [badger API](https://github.com/dgraph-io/badger/blob/release/v3.2103/txn.go#L761) said:

```go
// When you create a new transaction, it is absolutely essential to call
// Discard(). This should be done irrespective of what the update param is set
// to. Commit API internally runs Discard, but running it twice wouldn't cause
// any issues.
//
//  txn := db.NewTransaction(false)
//  defer txn.Discard()
//  // Call various APIs.
func (db *DB) NewTransaction(update bool) *Txn {
	return db.newTransaction(update, false)
}
```

But badgerengine doesn't call `txn.Discard()` before `txn.Commit`

![image](https://user-images.githubusercontent.com/26020263/160361260-ba85e5f1-6f24-4b36-94ad-dfe01716d74c.png)

### What did you expect to see?

Release disk space after I delete data and badger gc finish.

### What did you see instead?

genjidb with badgerdb consume too much disk space and badger gc doesn't work.

crazycs520:
https://github.com/genjidb/genji/pull/455/commits/41729dc2ed3e68a54b1b867f9f79125d92f0d7c9

I have bug-fix PR, but look like `badgerengine` has been removed.

asdine:
This is solved by calling Rollback after Commit, the same way as Badger:

```go
defer tx.Rollback()
...
return tx.Commit()
```

The issue is that `Rollback` documentation is not clear enough, it must definitely be improved.

> but look like badgerengine has been removed.

Yes, It has been removed for several reasons:

- Badger doesn't seem to be actively developed for the moment: https://github.com/dgraph-io/badger/graphs/commit-activity
- Dgraph has been having some financial issues these past months so it's not clear if Badger is a priority
- Genji is gonna come with it's own MVCC implementation, and Pebble seems to be the perfect choice for the backend

Genji is still a work in progress, so these kind of architectural changes are expected to happen until it reaches v1 unfortunately.

asdine:
Closing this as Badger is no longer supported and the new release will be published soon

PR #455, opened 2022-03-28
fix the issue of badgerengine doesn't release disk space since badger gc doesn't work

crazycs520:
close: https://github.com/genjidb/genji/issues/454


asdine:
@crazycs520 I'm closing this issue, as Badger is no longer supported and this behavior is handled with Pebble.
Thanks!

Issue #48, opened 2019-12-12
Scaling / HA using Raft / Multiraft


joeblew99:
Do any of the dB drivers have any scaling aspects ?

I would like to use genji in the traditional embedded scenario on desktops but also in servers. So for server usage I need a way to make it use Raft.

I saw bitcask and bitraft references in the issues. Looks like a promising candidate but it's not integration into genji yet.
@prologic: happy to know the status.

I think that is a badger Multiraft implementation on GitHub.

Please let me know blockers so I can decide if we should help do contribution

joeblew99:
https://github.com/BBVA/raft-badger

Looks promising..

winwisely99:
This is a much preferred implementation of a raft framework.
It also supports the new pebble store 

https://github.com/lni/dragonboat

protosam:
To have a scalable system with raft, even multiraft, isn't very straight forward. There is quite a bit to consider. [MIT 6.824: Distributed Systems](https://www.youtube.com/@6.824) can abridge some of this knowledge for anyone not in the know right now.

What should the FSM do? If it is only doing GET/PUT/DEL operations, the database is only highly available. This wont help with scaling data distribution or read/write scaling. It would only be useful as a locking system, from which coordination can be done in addition to.

Assuming that multiraft is used, one group would still need to maintain coordination of other groups. An entire system schema is necessary for that to work effectively. 

The common problems to solve for multiraft are:
- how to handle coordination? (the answer is always some form of distributed locking and placement driver)
- scaling reads and writes (each group is fully replicated, think write amplification)
- handling transactions (sometimes these are partitioned to their own raft groups OR timestamping with 2Phase commits on top of MVCC)

PR #460, opened 2022-07-01
Refactor kv.Store to interface

morlay:
Make `kv.Store` as interface 

We may replace to any other kv storage or distributed/remote kv for https://github.com/genjidb/genji/issues/48


asdine:
Hey @morlay thanks for submitting this PR and for putting so much effort!
Unfortunately, I can't accept this change because of the following reasons:

- Genji just went from having multiple store implementations (see [v0.14](https://github.com/genjidb/genji/tree/v0.14.1/engine), we had 3 different implementations) to a single one. Maintaining multiple implementations is a LOT of work, I have been trying it for years and I believe is not worth it. [Latest release](https://github.com/genjidb/genji/releases/tag/v0.15.0) focuses on a single backend for exactly this reason.
 - Most kv-stores come with their own transactions, whereas Genji implements its own transaction mechanism which relies a lot on Pebble primitives, which are hard to find in other kv-stores.
- Even if we decide to merge this PR, you will quickly realize how hard it is to implement another kv-store: Most of them don't allow you to pass your own custom [Compare](https://github.com/genjidb/genji/blob/main/internal/encoding/helpers.go#L100) function. Keys in Genji are not lexicographically sorted, but require to be in a certain order when we iterate on them. Most kv-stores use `bytes.Compare` to determine which key is greater than the other, and this won't work with the [binary](https://github.com/genjidb/genji/blob/main/internal/tree/key.go#L26-L47) representation of keys.

Basically, focusing on a single kv-store makes everything *much* easier to maintain, optimize and control.

They are plenty of things to do in Genji, and I would be more than happy to discuss about ideas in issues, so please don't hesitate to open one if you want any change.

morlay:
@asdine 

I just want to extends Genji scalable like the issue metioned https://github.com/genjidb/genji/issues/48.
`kv.Store` as interface could reach this goal, with implememting something like `raft pebble`.

Then Genji will be a Progressive Storage, from embed to distributed 😄 .

asdine:
@morlay I don't think this is the right way to make Genji distributed. This is a too big of a change to be done without proper discussion and planning sorry.
At the moment, the focus is mainly on adding SQL features and reaching stability. Even if we added raft to Genji, it wouldn't be very useful as it doesn't yet have enough SQL features.
I'm closing this PR but I appreciate the time you spent on this.

PR #461, opened 2022-07-01
add .restore command

sashaostrikov:
Restore command works exactly like `genji restore` ~and restarts the shell to open the database which was restored from the dump.~ and applies given dump to currently open database

~@asdine I also thought about adding a flag to decide whether:~
~1) we want to restart the shell and go to restored database~
~or~
~2) stay in current database.~

~But probably I am trying to overthink this and option 1 (which is currently implemented) is fine~

### DIFF
--- a/cmd/genji/commands/restore.go
+++ b/cmd/genji/commands/restore.go
@@ -1,9 +1,6 @@
 package commands
 
 import (
-	"io/ioutil"
-	"os"
-
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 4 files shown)


asdine:
Thanks @sashaostrikov !

I don't think we need to reload the database actually, the role of restore is just to apply the given queries to the **current** database.
We should basically:
- open the file
- iterate over all queries and execute them
- close the file

sashaostrikov:
@asdine PTAL, I've simplified the restore logic

asdine (approved):
Thanks!

PR #462, opened 2022-07-01
remove unused contexts from cmd/genji

sashaostrikov:
Just a cleanup of unused `context.Context` input parameters


asdine (approved):
Thanks!

PR #477, opened 2022-07-15
select: return error if field doesn't exist on tables with strict schema

### LINKED ISSUE #465: SELECT: If the schema is strict and a projected field doesn't exist, we should return an error
```sql
genji> create table foo(a int);
genji> insert into foo (a) values (1);
genji> select a, b from foo;
{
  "a": 1,
  "b": null
}
```

An error should be returned instead

The detection must be done during the [preparation](https://github.com/genjidb/genji/blob/818eee90f94ded67a7c4a2ccfc7fcd8a71d82803/internal/query/statement/select.go#L23) phase, before executing the query.

tzzed:
This PR fixes #465.

### DIFF
--- a/internal/query/statement/explain_test.go
+++ b/internal/query/statement/explain_test.go
@@ -4,9 +4,10 @@ import (
 	"encoding/json"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/genjidb/genji"
 	"github.com/genjidb/genji/internal/testutil/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func TestExplainStmt(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/query/statement/explain_test.go
```diff
@@ -20,18 +21,19 @@ func TestExplainStmt(t *testing.T) {
 		{"EXPLAIN SELECT * FROM test", false, `"table.Scan(\"test\")"`},
 		{"EXPLAIN SELECT *, a FROM test", false, `"table.Scan(\"test\") | docs.Project(*, a)"`},
 		{"EXPLAIN SELECT a + 1 FROM test", false, `"table.Scan(\"test\") | docs.Project(a + 1)"`},
-		{"EXPLAIN SELECT a + 1 FROM test WHERE c > 10", false, `"table.Scan(\"test\") | docs.Filter(c > 10) | docs.Project(a + 1)"`},
-		{"EXPLAIN SELECT a + 1 FROM test WHERE c > 10 AND d > 20", false, `"table.Scan(\"test\") | docs.Filter(c > 10) | docs.Filter(d > 20) | docs.Project(a + 1)"`},
-		{"EXPLAIN SELECT a + 1 FROM test WHERE c > 10 OR d > 20", false, `"table.Scan(\"test\") | docs.Filter(c > 10 OR d > 20) | docs.Project(a + 1)"`},
-		{"EXPLAIN SELECT a + 1 FROM test WHERE c IN [1 + 1, 2 + 2]", false, `"table.Scan(\"test\") | docs.Filter(c IN [2, 4]) | docs.Project(a + 1)"`},
+		{"EXPLAIN SELECT a + 1 FROM test WHERE c > 10", true, ``},
```
asdine, line 24: These tests were written to test something specific, we should not disable them by expecting them to fail.
We should instead either:
- change the declaration of the table so that it's not a strict table anymore
- change the path `c` with something that is already declared
- declare missing paths in the table declaration

asdine:
Closing this as the codebase have changed quite a bit, and this is no longer neeed. Thanks @tzzed !

PR #479, opened 2022-07-29
clean up some docs and code, including:

sashaostrikov:
- typo fixes
- docstring format fixes
- unused variables fixes
- introducing first usages of testutil.ParseDocumentPaths function, which was previously unused

I was reading the code and decided to add some changes to make it a little bit better :)

Test plan:
Existing tests should pass


asdine (approved):
Thanks! ❤️

PR #482, opened 2022-10-10
Add LEN builtin function

### LINKED ISSUE #474: Add `len` function for array type
### Proposal

Currently, it’s not possible to compute the length of array type in Genji

Darkclainer:
Fixes #474

I have few concerns about this PR:

* @asdine provided examples of using LEN with columns in tables. But this PR doesn't contain such tests, because tests will be required outside of `expr` package (or import of another parser will be needed). Anyway, I don't see anything bad in such test, but I don't known codebase well enough to decide is it worth it.
* Maybe error message can be improved.
* LEN function is quite expensive for ARRAY and DOCUMENT because it need to loop over elements. Probably this can be problem, but out of scope of  the mentioned issue and this PR.

### DIFF
--- a/internal/expr/functions/builtins.go
+++ b/internal/expr/functions/builtins.go
@@ -60,6 +60,13 @@ var builtinFunctions = Definitions{
 			return &Avg{Expr: args[0]}, nil
 		},
 	},
+	"len": &definition{
+		name:  "len",
+		arity: 1,
+		constructorFn: func(args ...expr.Expr) (expr.Function, error) {
+			return &Len{Expr: args[0]}, nil
+		},
+	},
 }
 
 // BuiltinDefinitions returns a map of builtin functions.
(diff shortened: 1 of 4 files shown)


internal/expr/functions/builtins.go
```diff
@@ -652,3 +659,62 @@ func (s *AvgAggregator) Eval(_ *environment.Environment) (types.Value, error) {
...
+
+// Eval extracts the average value from the given document and returns it.
+func (s *Len) Eval(env *environment.Environment) (types.Value, error) {
+	val, err := s.Expr.Eval(env)
+	if err != nil {
+		return nil, err
+	}
+	var length int
+	switch v := val.V().(type) {
+	case string:
```
asdine, line 678: We should always use the `v.Type()` method and compare with the types defines in the types package (see [here](https://github.com/genjidb/genji/blob/487d4b4eec9331c86b4ead7d30aa4eb2e8b1e7f1/types/types.go#L22-L41)). The internal representation of values can sometimes differ (strings stored as []byte for example).

Example:

```go
switch t := val.Type() {
  case types.TextValue:
    // ...
  case types.DocumentValue:
   // ...
}
```

internal/expr/functions/builtins.go
```diff
@@ -652,3 +659,62 @@ func (s *AvgAggregator) Eval(_ *environment.Environment) (types.Value, error) {
...
+	if !ok {
+		return false
+	}
+
+	return expr.Equal(s.Expr, o.Expr)
+}
+
+func (s *Len) Params() []expr.Expr { return []expr.Expr{s.Expr} }
+
+// String returns the alias if non-zero, otherwise it returns a string representation
```
asdine, line 716: This docstring is outdated, perhaps something like this `String returns the literal representation of len.`

asdine (commented):
Great job @Darkclainer !

> @asdine provided examples of using LEN with columns in tables. But this PR doesn't contain such tests, because tests will be required outside of expr package (or import of another parser will be needed). Anyway, I don't see anything bad in such test, but I don't known codebase well enough to decide is it worth it.

Actually, you can add tests in the [sqltests](https://github.com/genjidb/genji/tree/main/sqltests) directory. It is used to test complete sql queries and declare the expected output.
Here is an [example](https://github.com/genjidb/genji/blob/main/sqltests/SELECT/distinct.sql).

Here is what I suggest:
- add a file in `sqltests/SELECT/len.sql`
- move your tests in that file and adapt them to use a complete SELECT statement
- add the missing examples defined in the issue

> Maybe error message can be improved.

Do you have an example?

> LEN function is quite expensive for ARRAY and DOCUMENT because it need to loop over elements. Probably this can be problem, but out of scope of the mentioned issue and this PR.

It depends on the ARRAY and DOCUMENT implementation, but yeah it might be a problem for big ones.
But this is an optimization that should be made in [document.Length](https://github.com/genjidb/genji/blob/487d4b4eec9331c86b4ead7d30aa4eb2e8b1e7f1/document/document.go#L44) and [document.ArrayLength](https://github.com/genjidb/genji/blob/487d4b4eec9331c86b4ead7d30aa4eb2e8b1e7f1/document/array.go#L10) in the future. This PR can be merged without that optimization.

Darkclainer:
Thank you for review, @asdine!

> Actually, you can add tests in the [sqltests](https://github.com/genjidb/genji/tree/main/sqltests) directory. It is used to test complete sql queries and declare the expected output. Here is an [example](https://github.com/genjidb/genji/blob/main/sqltests/SELECT/distinct.sql).
> 
> Here is what I suggest:
> 
> * add a file in `sqltests/SELECT/len.sql`
> * move your tests in that file and adapt them to use a complete SELECT statement
> * add the missing examples defined in the issue

Did it, now it seems more complete.

> > Maybe error message can be improved.
> 
> Do you have an example?
> 

I do not. Actually, for me, any error in patched code seems like serious bug and should never be triggered, except for the first Eval that probably can fail. But may be you know situations when such cases can occur.

internal/expr/functions/builtins.go
Darkclainer: Fixed it with your suggestion

internal/expr/functions/builtins.go
asdine: I'll keep an eye on this and find a way to make sure we never create non-string TEXT values

asdine (approved):
Thanks @Darkclainer !

PR #484, opened 2022-11-11
Add forgotten import in example

belanenko:
Add forgotten import in example

	"github.com/genjidb/genji/types"


asdine (approved):
Good catch, thanks!

Issue #453, opened 2022-03-10
StructScan panics when the interface parameter is not a struct


tzzed:
### What version of Genji are you using?

<pre>
$ genji version
Genji (devel)
Genji CLI (devel)

</pre>

### What did you do?
StructScan panics when the api contract is not respected, it will be very good an error is returned instead of panic.

```go
        var b int
	d := document.NewFieldBuffer().Add("a", types.NewIntegerValue(10))
        err := document.StructScan(d, &b)
	assert.Error(t, err)
```

### What did you expect to see?
An error message.
```go
target must be pointer to struct
```

### What did you see instead?
```
panic: reflect: call of reflect.Value.NumField on int Value [recovered]
	panic: reflect: call of reflect.Value.NumField on int Value
```

asdine:
Fixed by #487

PR #487, opened 2022-12-17
StructScan check interface parameter is a pointer to a struct

abramlab:
Added check that an interface parameter is a pointer to a structure in StructScan.

Refers #453


asdine (approved):
Thanks @abramlab (sorry for the late review)

Issue #444, opened 2022-02-11
import json file


tzzed:
### Proposal
As we can import csv file, it will be useful to import json file.
Even if we can use the genji command insert like this `cat foo.json | genji insert --db mydb -t foo`
```shell
genji>.import json foo.json foo
```

### Motivation
Inside the cli, we can need to import json files as we import csv files.

PR #488, opened 2022-12-18
Import json file

abramlab:
* Added support to import json files
 * Separated csv insert logic into InsertCSV function and moved it to "cmd/genji/dbutil/insert.go"

Refers #444


asdine:
Not needed anymore, thanks!

PR #504, opened 2023-08-06
Add schema validation for strict schema table in core select statements

Darkclainer:
This is early draft, please see #465 for discussion.


asdine:
Not needed anymore, thanks!

PR #510, opened 2023-11-17
Add LOWER builtin function

agonist:
Adds LOWER to builtins functions to format TEXT fields to lower-case.

For non TEXT fields it will return NULL.

Types that can be casted to TEXT will work if they're first explicitly casted.


asdine (approved):
Great job @agonist !

PR #512, opened 2023-11-17
add trim, ltrim, rtrim functions

agonist:
Add the trim function and its 2 variants ltrim rtrim.

### DIFF
--- a/internal/expr/functions/definition.go
+++ b/internal/expr/functions/definition.go
@@ -23,8 +23,8 @@ type Packages map[string]Definitions
 
 func DefaultPackages() Packages {
 	return Packages{
-		"":     BuiltinDefinitions(),
-		"math": MathFunctions(),
+		"":        BuiltinDefinitions(),
+		"math":    MathFunctions(),
 		"strings": StringsDefinitions(),
 	}
 }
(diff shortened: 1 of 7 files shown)


internal/expr/functions/definition.go
```diff
@@ -57,13 +57,20 @@ func (fd *definition) Name() string {
...
+
 	if len(args) != fd.arity {
 		return nil, fmt.Errorf("%s() takes %d argument(s), not %d", fd.name, fd.arity, len(args))
 	}
 	return fd.constructorFn(args...)
 }
 
 func (fd *definition) String() string {
+	if fd.arity < 0 {
+		fd.arity = 0
```
asdine, line 72: I wouldn't change the arity here. Instead we could use a local variable
```go
arity := fd.arity
if arity < 0 {
...
```

internal/expr/functions/strings.go
```diff
@@ -109,3 +130,72 @@ func (s *Upper) Params() []expr.Expr { return []expr.Expr{s.Expr} }
...
+		}
+		cutset = types.As[string](remove)
+	}
+
+	trimmed := s.TrimFunc(types.As[string](input), cutset)
+
+	return types.NewTextValue(trimmed), nil
+}
+
+func (s *Trim) IsEqual(other []expr.Expr) bool {
```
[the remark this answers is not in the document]
agonist: fixed

internal/expr/functions/definition.go
agonist: fixed

asdine (approved):
👏🏼

PR #513, opened 2023-11-21
Add coalesce function

### LINKED ISSUE #509: Functions: COALESCE
Add the COALESCE function.

Follow the [PostgreSQL](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-COALESCE-NVL-IFNULL) definition.

yaziine:
This PR fixes #509 by adding support to `COALESCE` method.
Unlike PosgreSQL, our implementation doesn't fail if the arguments aren't of the same type.
However, an error is returned if there is no argument at all.

### DIFF
--- a/internal/expr/functions/definition.go
+++ b/internal/expr/functions/definition.go
@@ -60,10 +60,6 @@ func (fd *definition) Name() string {
 }
 
 func (fd *definition) Function(args ...expr.Expr) (expr.Function, error) {
-	//if fd.arity == -1 {
-	//	return fd.constructorFn(args...)
-	//}
-
 	if fd.arity != UNLIMITED && (len(args) != fd.arity) {
 		return nil, fmt.Errorf("%s() takes %d argument(s), not %d", fd.name, fd.arity, len(args))
 	}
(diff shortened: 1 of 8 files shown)


internal/sql/parser/parser.go
```diff
@@ -268,6 +268,9 @@ func (p *Parser) parseTokens(tokens ...scanner.Token) error {
 // present, it unscans and return false. If the first is present, all the others
 // must be parsed otherwise an error is returned.
 func (p *Parser) parseOptional(tokens ...scanner.Token) (bool, error) {
+	if len(tokens) == 0 {
```
yaziine, line 271: This fixes a potential panic if tokens are empty.

internal/expr/functions/definition.go
```diff
@@ -7,6 +7,9 @@ import (
 	"github.com/genjidb/genji/internal/expr"
 )
 
+// UNLIMITED represents an unlimited number of arguments.
+const UNLIMITED = -1
```
asdine, line 11: Full uppercase is not really idiomatic. Consider changing this to something like`variadicArity`

internal/sql/parser/parser.go
asdine: Panics are good in these cases, we don't want to use this method with empty arguments

asdine (approved):
👏🏼

PR #514, opened 2023-11-24
Add random function

agonist:


internal/expr/functions/math.go
```diff
@@ -164,3 +167,18 @@ var atan2 = &ScalarDefinition{
 		return types.NewDoubleValue(res), nil
 	},
 }
+
+var random = &ScalarDefinition{
+	name:  "random",
+	arity: 0,
+	callFn: func(args ...types.Value) (types.Value, error) {
+		src := rand.NewSource(time.Now().UnixNano())
```
asdine, line 175: Actually, all you need to do is `rand.Int63()`. It uses the default source, which is already [seeded](https://stackoverflow.com/questions/12321133/how-to-properly-seed-random-number-generator/74077488#74077488) since Go 1.20.

agonist: oh well. that's nice

PR #532, opened 2024-11-01
Fix the quickstart example so it actually works

samjewell:
All these changes were needed for this quickstart example to run through without crashing


samjewell:
@asdine Can I have a review on this one please?

asdine (approved):
Thanks!

PR #541, opened 2025-08-15
Bump pebble

cgruber:


asdine (commented):
@cgruber Thanks! The CI fails because of go.mod issue. You might need to call `go mod tidy` I guess

cgruber:
Ah crap - I was wondering - i didn't notice it was a separate mod. On it.

cgruber:
My bad. I didn't run the tests for /cmd locally.

asdine:
You can use `make test`

asdine:
Haha no worries, plus there's no documentation on how to run them so one can only guess

asdine:
Keep in mind that the project is a WIP and it's taking a different direction than what's in the `main` README, so I would advise (perhaps a bit late) to not use it in production for now (unless you're fine with using only the in-memory version and some breaking changes in the future). cf [dev](https://github.com/chaisql/chai/tree/dev) branch

asdine (approved):
Build is green! Thanks @cgruber !

cgruber:
Awesome. Yeah - I'm only using it for a fake implementation used for unit testing, and only in-memory. If you take it in a different direction I hope you keep that use-case at least. I had this in Java/Kotlin (in-memory sql db) and it's just so handy to avoid long integration tests.

asdine:
Yeah the new direction is mostly for the SQL API (it will be Postgres compatible) and the on-disk file format. I'll keep the in-memory part, CI is a big use case for this library.

